PR #95, opened 2020-09-29
[ISSUE-111] Add system disk support

AnastasiaShemyakinskaya:
## Purpose
* Add new field in drive type
* Check root mount during updateDriveCRs
* Set system lvg location as system disk
Solves https://github.com/dell/csi-baremetal/issues/111

## PR checklist
- [ ] Add link to the issue
- [x] Choose PR label
- [x] New unit tests added
- [ ] All TODOs are linked with issues
- [x] PR validation passed
- [x] Custom CI passed


pkg/base/k8s/kubeclient.go
```diff
@@ -345,6 +345,24 @@ func (k *KubeClient) GetNodes(ctx context.Context) ([]coreV1.Node, error) {
 	return nodes.Items, nil
 }
 
+// GetSystemDriveUUID returns system drive uuid
+// Receives golang context
+// Returns return string - drive uuid, error
+func (k *KubeClient) GetSystemDriveUUID(ctx context.Context, node string) string {
```
ykulazhenkov, line 351: How we handle situation when OS installed on top of RAID and we have multiple system drives?

AnastasiaShemyakinskaya: I didn't implement support for multiple system disks. I'll think about it

AnastasiaShemyakinskaya: @ykulazhenkov I assume we don't support lvg on multiple system disks yet. As I see we create LVG for one system disk. Does it make sense to support this case first?

ykulazhenkov: If I understand you correctly, you talk about this use case:
/dev/sda, /dev/sdb -> LVM system-VG - > LVM root-lv (used for /)- 2 physical devices in VG, LV with root fs in this VG
I think we should support this use case. Drives /dev/sda and /dev/sdb should be marked as system drives.

Originally I was talking about another use cases with non-LVM RAID:
/dev/sda, /dev/sdb -> RAID /dev/md1 -> /  OR
/dev/sda, /dev/sdb -> RAID /dev/md1 -> LVM system-VG -> LVM root-lv (used for /)
We should improve system drives detection to handle this scenarios.

AnastasiaShemyakinskaya: Okay, I've got you

pkg/base/k8s/kubeclient.go
```diff
@@ -345,6 +345,25 @@ func (k *KubeClient) GetNodes(ctx context.Context) ([]coreV1.Node, error) {
 	return nodes.Items, nil
 }
 
+// GetSystemDriveUUID returns system drive uuid
+// Receives golang context
+// Returns return string - drive uuid, error
```
ykulazhenkov, line 350: Please, fix types in docstring

pkg/common/ac_operations.go
```diff
@@ -278,15 +279,21 @@ func (a *ACOperationsImpl) tryToFindAC(acs []*accrd.AvailableCapacity, storageCl
 	var (
 		allocatedCapacity int64 = math.MaxInt64
 		foundAC           *accrd.AvailableCapacity
+		driveUUID         = make([]string, 0)
 	)
 	for _, ac := range acs {
+		// Available capacity with system disk location won't be allocated
+		if len(driveUUID) == 0 {
+			driveUUID = a.k8sClient.GetSystemDriveUUID(context.Background(), ac.Spec.NodeId)
```
ykulazhenkov, line 287: Maybe we can try to move this code outside of the loop? Can we retrieve system drives for all nodes with single request?

pkg/common/ac_operations.go
```diff
@@ -278,15 +279,21 @@ func (a *ACOperationsImpl) tryToFindAC(acs []*accrd.AvailableCapacity, storageCl
...
+		}
+		driveUUID = append(driveUUID, base.SystemDriveAsLocation)
 		if storageClass != "" {
-			if ac.Spec.Size < allocatedCapacity && ac.Spec.Size >= requiredBytes && ac.Spec.StorageClass == storageClass {
+			if ac.Spec.Size < allocatedCapacity && ac.Spec.Size >= requiredBytes && ac.Spec.StorageClass == storageClass && !util.IsDriveSystem(ac.Spec.Location, driveUUID) {
```
ykulazhenkov, line 291: Do we need to always ignore system drives? If so, maybe we can do IsDriveSystem check earlier?
 ```
if util.IsDriveSystem(ac.Spec.Location, driveUUID) {
   continue
}
```

File: pkg/base/k8s/kubeclient.go [hunk 5]
```diff
@@ -345,6 +345,25 @@ func (k *KubeClient) GetNodes(ctx context.Context) ([]coreV1.Node, error) {
...
+// Returns return slice of string - system drives uuid
+func (k *KubeClient) GetSystemDriveUUID(ctx context.Context) []string {
+	ll := k.log.WithField("method", "GetSystemDriveUUID")
+	var driveList drivecrd.DriveList
+	if err := k.ReadList(ctx, &driveList); err != nil {
+		ll.Errorf("Failed to get system disk uuid, error: %v", err)
+		return nil
+	}
+	driveUUID := make([]string, 0)
+	for _, drive := range driveList.Items {
```
mishoyama, line 359: Do we need to return error hen drive not found?

pkg/common/ac_operations.go
```diff
@@ -278,8 +279,17 @@ func (a *ACOperationsImpl) tryToFindAC(acs []*accrd.AvailableCapacity, storageCl
 	var (
 		allocatedCapacity int64 = math.MaxInt64
 		foundAC           *accrd.AvailableCapacity
+		driveUUID         = make([]string, 0)
 	)
 	for _, ac := range acs {
+		// Available capacity with system disk location won't be allocated
```
mishoyama, line 285: When system drive has free space nothing prevents us of using it.

pkg/common/ac_operations.go
```diff
@@ -278,8 +279,17 @@ func (a *ACOperationsImpl) tryToFindAC(acs []*accrd.AvailableCapacity, storageCl
...
+		if len(driveUUID) == 0 {
+			driveUUID = a.k8sClient.GetSystemDriveUUID(context.Background())
+			driveUUID = append(driveUUID, base.SystemDriveAsLocation)
+		}
+		if util.IsDriveSystem(ac.Spec.Location, driveUUID) {
```
mishoyama, line 290: This code looks weird. You just get UUID of a system drive and later check whether is system or not.

pkg/node/volumemgr.go
```diff
@@ -996,3 +1019,29 @@ func prepareDriveDescription(drive *drivecrd.Drive) string {
...
+	if err != nil {
+		return false, err
+	}
+	return m.isRootMountpoint(devices), nil
+}
+
+// isRootMountpoint check whether devices has root mountpoint
+// Parameters: BlockDevice from lsblk output
+// Returns true if device has root mountpoint, false in opposite
+func (m *VolumeManager) isRootMountpoint(dev []lsblk.BlockDevice) bool {
```
mishoyama, line 1037: Do we need to check for root only? What about others?

mishoyama (changes requested):

pkg/drivemgr/basemgr/basemgr.go
```diff
@@ -56,19 +56,19 @@ func (mgr LinuxUtilManager) GetDrivesList() ([]*api.Drive, error) {
 	return devices, nil
 }
 
-//NewLinuxUtilManager is a constructor LinuxUtilManager
-func NewLinuxUtilManager(exec command.CmdExecutor, logger *logrus.Logger) *LinuxUtilManager {
-	return &LinuxUtilManager{
+//NewBaseManager is a constructor BaseManager
+func NewBaseManager(exec command.CmdExecutor, logger *logrus.Logger) *BaseManager {
```
painhardcore, line 60: Maybe we should use "New" instead of "NewBaseManager" since it's the only one constructor in the package.
`basemgr.New(....)` looks short and concise.

pkg/base/k8s/kubeclient.go
AnastasiaShemyakinskaya: Do you mean return an error if there are no system drives were found in `driveList`? In this case we can log error and return the empty slice.

mishoyama: Yes, please. It will help with the troubleshooting.

pkg/base/k8s/kubeclient.go
AnastasiaShemyakinskaya: Done

pkg/common/ac_operations.go
AnastasiaShemyakinskaya: I removed this code due to this comment https://github.com/dell/csi-baremetal/pull/95#discussion_r498657251

pkg/common/ac_operations.go
AnastasiaShemyakinskaya: Same

pkg/base/k8s/kubeclient.go
```diff
@@ -345,6 +345,28 @@ func (k *KubeClient) GetNodes(ctx context.Context) ([]coreV1.Node, error) {
 	return nodes.Items, nil
 }
 
+// GetSystemDriveUUID returns system drives uuid
+// Receives golang context
+// Returns return slice of string - system drives uuid
+func (k *KubeClient) GetSystemDriveUUID(ctx context.Context) []string {
```
rususlasan, line 351: Since method returns a slice of UUIDs let's rename it to: `GetSystemDriveUUIDs`

pkg/base/k8s/kubeclient.go
AnastasiaShemyakinskaya: Fixed

pkg/common/ac_operations.go
AnastasiaShemyakinskaya: Fixed

pkg/base/k8s/kubeclient.go
```diff
@@ -345,6 +345,28 @@ func (k *KubeClient) GetNodes(ctx context.Context) ([]coreV1.Node, error) {
...
+	ll := k.log.WithField("method", "GetSystemDriveUUID")
+	var driveList drivecrd.DriveList
+	if err := k.ReadList(ctx, &driveList); err != nil {
+		ll.Errorf("Failed to get system disk uuid, error: %v", err)
```
rususlasan, line 355: Here you read not only system drives but all drives.

pkg/common/ac_operations.go
AnastasiaShemyakinskaya: Fixed

pkg/lvm/lvgcontroller.go
```diff
@@ -124,11 +124,14 @@ func (c *LVGController) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 
 			if len(lvg.Spec.Locations) == 0 {
 				ll.Warn("Location fields is empty")
-			} else if lvg.Spec.Locations[0] != base.SystemDriveAsLocation {
-				// cleanup LVM artifacts
-				if err := c.removeLVGArtifacts(lvg.Name); err != nil {
-					ll.Errorf("Unable to cleanup LVM artifacts: %v", err)
-					return ctrl.Result{}, err
+			} else {
+				driveUUID := append(c.k8sClient.GetSystemDriveUUID(context.Background()), base.SystemDriveAsLocation)
```
rususlasan, line 128: here you get slice, I think `driveUUIDs` is more clear

pkg/base/util/drive_helper.go
```diff
@@ -0,0 +1,29 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package util
+
+// IsDriveSystem check if given uuid is in systemDisks array
+// Receive: uuid of drive, array of system disks uuid
+// Return: true if uuid is in systemDisks, otherwise false
+func IsDriveSystem(uuid string, systemDisks []string) bool {
```
rususlasan, line 22: Why do you place that method in separate file? You have method `isDriveSystem` in volume manager that does real work. But here you just check that slice contains string. We have method that check whether slice contains string or not. Let's use it.

pkg/node/volumemgr.go
rususlasan, line 1037: since dev it is a slice (not a single object) could you please rename dev to devs

mishoyama (approved):

pkg/common/volume_operations.go
```diff
@@ -346,8 +346,9 @@ func (vo *VolumeOperationsImpl) deleteLVGIfVolumesNotExistOrUpdate(lvg *lvgcrd.L
 		"volumeID": volID,
 	})
 
+	driveUUID := append(vo.k8sClient.GetSystemDriveUUIDs(context.Background()), base.SystemDriveAsLocation)
```
rususlasan, line 349: driveUUID it is a slice, let's use more meaningful name: `driveUUIDs`. When  read name like `driveUUID` I suppose that this variable stores something singular, like string.

pkg/base/k8s/kubeclient.go
```diff
@@ -345,6 +345,28 @@ func (k *KubeClient) GetNodes(ctx context.Context) ([]coreV1.Node, error) {
 	return nodes.Items, nil
 }
 
+// GetSystemDriveUUIDs returns system drives uuid
+// Receives golang context
+// Returns return slice of string - system drives uuids
+func (k *KubeClient) GetSystemDriveUUIDs(ctx context.Context) []string {
```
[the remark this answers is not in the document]
AnastasiaShemyakinskaya: This method is used in lvg controller, which doesn't have CR helper in structure. Should we extend this struct and add cr_helper only for one method? The same situation with volume_operations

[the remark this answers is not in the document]
rususlasan: Ok, let's leave this method in kubeclient, but what about context? I don't see why do we need context as a parameter for that method.

rususlasan (approved):

ykulazhenkov (approved):

Issue #77, opened 2020-09-25
Implement prioritize method in scheduler


rususlasan:
The scheduler extender should handle prioritize requests from kube-scheduler and score nodes based on the amount of created volumes on the node.
Currently prioritize is not Implemeneted, this may cause a situation overloading one on the cluster.
I.e. we have a cluster with a high amount of resources, all nodes will satisfy our filtering phase for N pods and they can be scheduled on 1 node.
The prioritize phase should rank nodes based on their volumes to distribute scheduling evenly

PR #104, opened 2020-10-01
Prioritize method in the extender

painhardcore:
## Purpose
Since we have our scheduler extender, we can set priorities for the nodes. We need to try equally schedule across all nodes.
Scoring solution based on the count of volumes.
A smaller number of volumes get a higher rank if a node has no volume - gets a big number to have higher priority.
Solves #77 
## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [x] New unit tests added
- [x] All TODOs are linked with issues
- [x] All comments are resolved
- [x] PR validation passed
- [x] Custom CI passed

### DIFF
--- a/pkg/scheduler/extender/extender.go
+++ b/pkg/scheduler/extender/extender.go
@@ -417,7 +417,7 @@ func (e *Extender) createACRs(ctx context.Context, nodeVolumeACMap map[string]ma
 	return createErr
 }
 
-// searchSuitableACs searches the most appropriate AC for each volume from volumes in scACMap
+
 func (e *Extender) score(nodes []coreV1.Node) ([]schedulerapi.HostPriority, error) {
 	ll := e.logger.WithFields(logrus.Fields{
 		"method": "score",
@@ -480,6 +480,7 @@ func nodePrioritize(nodeMapping map[string][]volcrd.Volume) (map[string]int, int
 	return nrank, maxCount
 }
 
+// searchSuitableACs searches the most appropriate AC for each volume from volumes in scACMap
 // scACMap - map that represents available capacities and has next structure: map[StorageClass][AC.Name]*AC
 func (e *Extender) searchSuitableACs(scACMap map[string]map[string]*accrd.AvailableCapacity,
 	sc string, volumes []*genV1.Volume) map[*genV1.Volume]*accrd.AvailableCapacity { // list based on which ACR will be created
--- a/pkg/scheduler/extender/extender.go
+++ b/pkg/scheduler/extender/extender.go
@@ -480,7 +480,6 @@ func nodePrioritize(nodeMapping map[string][]volcrd.Volume) (map[string]int, int
 	return nrank, maxCount
 }
 
-// isACsMatchVolumeRequests checks whether volumes suite with storage class sc could be provisioned based on available capacities
 // scACMap - map that represents available capacities and has next structure: map[StorageClass][AC.Name]*AC
 func (e *Extender) searchSuitableACs(scACMap map[string]map[string]*accrd.AvailableCapacity,
 	sc string, volumes []*genV1.Volume) map[*genV1.Volume]*accrd.AvailableCapacity { // list based on which ACR will be created
--- a/pkg/scheduler/extender/extender.go
+++ b/pkg/scheduler/extender/extender.go
@@ -417,7 +417,6 @@ func (e *Extender) createACRs(ctx context.Context, nodeVolumeACMap map[string]ma
 	return createErr
 }
 
-
 func (e *Extender) score(nodes []coreV1.Node) ([]schedulerapi.HostPriority, error) {
 	ll := e.logger.WithFields(logrus.Fields{
 		"method": "score",
--- a/pkg/scheduler/extender/extender.go
+++ b/pkg/scheduler/extender/extender.go
@@ -30,7 +30,7 @@ import (
 	coreV1 "k8s.io/api/core/v1"
 	storageV1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
-	schedulerapi "k8s.io/kubernetes/pkg/scheduler/api"
+	schedulerapi "k8s.io/kubernetes/pkg/scheduler/api/v1"
 	k8sCl "sigs.k8s.io/controller-runtime/pkg/client"
 
 	genV1 "github.com/dell/csi-baremetal/api/generated/v1"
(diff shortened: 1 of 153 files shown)


mishoyama (changes requested):
Please add more details to the issue and PR descriptions about what we are trying to archive here.

pkg/scheduler/extender/extender.go
```diff
@@ -135,9 +137,41 @@ func (e *Extender) FilterHandler(w http.ResponseWriter, req *http.Request) {
...
+	)
+
+	if err := json.NewDecoder(req.Body).Decode(&extenderArgs); err != nil {
+		ll.Errorf("Unable to decode request body: %v", err)
+		return
+	}
+
+	ll.Info("Scoring")
+
+	e.Lock()
```
mishoyama, line 162: Do we really need to have a lock here?

painhardcore: @mishoyama It's not 100% required here, but It will help with parallel requests, so ranking will be more precise.
Also, the lock will protect Kubernetes API from being spammed, If we do lots of parallel requests to this endpoint.

pkg/scheduler/extender/extender.go
```diff
@@ -135,9 +137,41 @@ func (e *Extender) FilterHandler(w http.ResponseWriter, req *http.Request) {
 }
 
 // PrioritizeHandler assigns scores to the nodes
```
mishoyama, line 139: Please add more detailed description.

pkg/scheduler/extender/extender.go
```diff
@@ -329,6 +363,81 @@ func (e *Extender) filter(ctx context.Context, nodes []coreV1.Node, volumes []*g
...
+		}
+		hostPriority = append(hostPriority, schedulerapi.HostPriority{
+			Host:  node.GetName(),
+			Score: rank,
+		})
+	}
+	return hostPriority, nil
+}
+
+func nodeVolumeCountMapping(vollist *volcrd.VolumeList) map[string][]volcrd.Volume {
```
mishoyama, line 400: Shouldn't you take into account only BM CSI volumes?

pkg/scheduler/extender/extender.go
```diff
@@ -329,6 +363,81 @@ func (e *Extender) filter(ctx context.Context, nodes []coreV1.Node, volumes []*g
...
+
+	priorityFromVolumes := nodePrioritize(sortedNoded)
+
+	ll.Debugf("nodes were ranked by their volumes %+v", priorityFromVolumes)
+	hostPriority := make([]schedulerapi.HostPriority, 0, len(nodes))
+	for _, node := range nodes {
+		// set the highest priority if node doesn't have any volumes
+		rank := math.MaxInt16
+		if r, ok := priorityFromVolumes[string(node.GetUID())]; ok {
+			rank = r
```
mishoyama, line 390: Does rank mean just a number of PV? Shouldn't it be calculated depending on other node's values? Will high rank cause issue with scores from CPU/memory and other extenders?

mishoyama (changes requested):

pkg/scheduler/extender/extender.go
```diff
@@ -329,6 +363,81 @@ func (e *Extender) filter(ctx context.Context, nodes []coreV1.Node, volumes []*g
...
 }
 
+func (e *Extender) score(nodes []coreV1.Node) ([]schedulerapi.HostPriority, error) {
+	ll := e.logger.WithFields(logrus.Fields{
+		"method": "score",
+	})
+
+	var volumeList = &volcrd.VolumeList{}
+	if err := e.k8sClient.ReadList(context.Background(), volumeList); err != nil {
+		err = fmt.Errorf("unable to read AvailableCapacity list: %v", err)
```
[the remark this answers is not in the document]
mishoyama: This is ok. We want to calc number of allocated volumes.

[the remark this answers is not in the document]
rususlasan: Yeah, but my comment points on error message.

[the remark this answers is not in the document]
painhardcore: @rususlasan thx will fix the error message.

pkg/scheduler/extender/extender.go
painhardcore: It's BM CSI volumes.

pkg/scheduler/extender/extender.go
painhardcore: It's based on the position in the sorted slice of nodes based on volume count.
Apparently, scoring is used only with extenders. So other extenders may affect the score of one of the nodes.
But I don't see how it could be an issue - scoring runs on the already filtered list of nodes and other extenders can have higher scores.

rususlasan (approved):

pkg/scheduler/extender/extender.go
mishoyama: Let's say you have _**one node with 100 volumes**_ and **_second with 1 volume only_**. Final scores will be 0 and 1 correspondingly. Other extenders will simply over score node with 1 volume I guess. An vice versa - when you have no volumes on the node(s) _MaxInt16_ score will impact placement decisions of others.

painhardcore: My intent was to force nodes without volume to have one. I can rewrite prioritize to make it more aligned on "something".
But there is no guarantee that other plugins won't over score us :)

mishoyama: The original intent was to have even distribution of the volumes across the nodes (vSAN uses term _balanced_). You can calсulate rank base on the following formula: _rank of node X = max number of volumes - number of volume on node X_. For example, if you have 4 nodes with 8, 7, 7 and 5 volumes, final ranks will be 8-8=0, 8-7=1, 8-7=1, 8-5=3. In this case we will prioritize the last node clearly.

painhardcore: Fixed.

mishoyama: You use total volume count for all nodes, but you need to find max number of volumes on a single node across nodes list from _extenderArgs.Nodes.Items_.

painhardcore: As a future improvement, we can  consider aligning priorities to MaxNodeScore. But it's a little bit confusing which value to use - [16.4](https://github.com/kubernetes/kubernetes/blob/224be7bdce5a9dd0c2fd0d46b83865648e2fe0ba/pkg/scheduler/framework/v1alpha1/interface.go#L74) - 10, [master](https://github.com/kubernetes/kubernetes/blob/1a66eb7b8aba8d5e66d855d6b8b31fdeab648778/pkg/scheduler/framework/v1alpha1/interface.go#L86) - 100.

painhardcore: Yep, right now I'm getting the max number of volumes on a single node, not all.

mishoyama (approved):

Issue #117, opened 2020-10-13
Create csi-baremetal.nodes CRD


AnastasiaShemyakinskaya:
CRD must contain:

1. Host name
2. IP address(es)
3. Node UUID

PR #121, opened 2020-10-13
[Issue-117] Create csi-baremetal.nodes CRD

AnastasiaShemyakinskaya:
## Purpose
* Add CSI node CRD
Solves https://github.com/dell/csi-baremetal/issues/117

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [ ] New unit tests added
- [x] PR validation is passed
- [x] Custom CI is passed

## Testing
_Link to custom CI build_

### DIFF
--- a/api/v1/types.proto
+++ b/api/v1/types.proto
@@ -64,8 +64,9 @@ message LogicalVolumeGroup {
     string Status = 6;
 }
 
-message CSINode {
+message CSIBMNode {
     string UUID = 1;
     repeated string IPs = 2;
     string Hostname = 3;
+    string Status = 4;
 }
--- a/api/v1/types.proto
+++ b/api/v1/types.proto
@@ -66,6 +66,6 @@ message LogicalVolumeGroup {
 
 message CSINode {
     string UUID = 1;
-    repeated string Ip = 2;
+    repeated string IPs = 2;
     string Hostname = 3;
 }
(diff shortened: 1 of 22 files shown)


api/v1/types.proto
```diff
@@ -63,3 +63,9 @@ message LogicalVolumeGroup {
     repeated string VolumeRefs = 5;
     string Status = 6;
 }
+
+message CSINode {
```
rususlasan, line 67: I think we need to add one more field here in which CSINode status will be stored:
`Status` - online/offline. 
Controller will use that field for cases when node will be replaced or accidentally goes down:
- some node goes down, e.g. NODE1
- controller set corresponding CR status to offline: CSINode.Spec.Status = "OFFLINE"
- new node is added to the cluster, new node has same hostname and ips that "old" node had
- controller detect that new node is corresponding to "old" node and change CSINOde.Spec.Status to "ONLINE" and restore node ID for new node based on value of CSINode.Spec.ID

rususlasan (approved):

api/v1/types.proto
mishoyama: Don't think that we need to track node status here.

rususlasan: What operator should to do when node goes down?

mishoyama (approved):

Issue #145, opened 2020-10-27
CSI should handle system disks on OpenShift nodes


AnastasiaShemyakinskaya:
If we deploy CSI on OpenShift, baremetal-csi-node pods have following output of lsblk inside the container:
```
root@baremetal-csi-node-c25sf:/# lsblk
NAME                                                                                     MAJ:MIN RM   SIZE RO TYPE MOUNTPOINT
sdb                                                                                        8:16   0 447.1G  0 disk 
|-sdb1                                                                                     8:17   0   384M  0 part 
|-sdb2                                                                                     8:18   0   127M  0 part 
|-sdb3                                                                                     8:19   0     1M  0 part 
`-sdb4                                                                                     8:20   0 446.6G  0 part 
  `-coreos-luks-root-nocrypt                                                             253:0    0 446.6G  0 dm   /var/lib/kubelet/pods/41eb6249-6f66-4fdb-81f4-87cc7564094c/volume-subpaths/etc/tuned/5
```
CSI uses lsblk output to discover system disks, so it needs to handle the case, when mountpoint != ` /var/lib/kubelet/pods/`

PR #146, opened 2020-10-27
[Issue-145] CSI should handle OpenShift system disks

AnastasiaShemyakinskaya:
## Purpose
Solves https://github.com/dell/csi-baremetal/issues/145

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [ ] New unit tests added
- [x] PR validation is passed
- [x] Custom CI is passed

## Testing
_Link to custom CI build_


charts/baremetal-csi-plugin/templates/baremetal-csi-node.yaml
```diff
@@ -233,6 +236,10 @@ spec:
         hostPath:
           path: /sys
           type: Directory
+      - name: host-root
```
rususlasan, line 239: Let's make it configurable and do not share host root by default.

AnastasiaShemyakinskaya: Done

pkg/base/constants.go
```diff
@@ -41,6 +41,10 @@ const (
 
 	// KubeletRootPath is the pods' path on the node
 	KubeletRootPath = "/var/lib/kubelet/pods"
+
+	// OpenShiftHostRootPath is root mount in pods for OpenShift cluster
+	OpenShiftHostRootPath = "/hostroot/sysroot"
```
mishoyama, line 46: I prefer to avoid having names of particular platforms in the code.

mishoyama (changes requested):

charts/baremetal-csi-plugin/values.yaml
```diff
@@ -4,6 +4,7 @@ global:
 
 env:
  test: false
+ openshift: false
```
mishoyama, line 7: please rename to `mountHostRoot` since this problem might occur on any platform.

mishoyama (changes requested):

pkg/base/constants.go
AnastasiaShemyakinskaya: Fixed

charts/baremetal-csi-plugin/values.yaml
AnastasiaShemyakinskaya: Fixed

mishoyama (approved):

rususlasan (approved):

painhardcore (approved):

Issue #168, opened 2020-11-23
Add LED API in drive manager


rususlasan:
Drive manager should be able to manipulate drive LED status. It is needed for disk replacement procedure for physical detect drive.
As part of that issue DriveManager interface should be expanded and default implementation of new methods should be added as well.

PR #171, opened 2020-11-24
[ISSUE-168] extend drivemgr interface

rususlasan:
## Purpose
Issue - https://github.com/dell/csi-baremetal/issues/168

- Extending DriveManager interface by adding method that should manipulate by drive's LED status.
- added default implementation of that method in all DriveManager implementations

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [ ] New unit tests added
- [x] PR validation is passed
- [ ] Custom CI is passed

## Testing
_Link to custom CI build_


AnastasiaShemyakinskaya (approved):

pkg/drivemgr/drivemgr_service.go
```diff
@@ -66,3 +65,14 @@ func (svc *DriveServiceServerImpl) GetDrivesList(ctx context.Context, req *api.D
 		Disks: drives,
 	}, nil
 }
+
+// Locate invokes DriveManager's Locate method for manipulation drive's LED state
+func (svc *DriveServiceServerImpl) Locate(ctx context.Context, in *api.DriveLocateRequest) (*api.DriveLocateResponse, error) {
```
painhardcore, line 70: Why do we need DriveLocateResponse?It's just a blank response?

rususlasan: Interface is defined in [drivesvc.proto](https://github.com/dell/csi-baremetal/blob/feature-ISSUE-168-add-led-interface/api/v1/drivemgrsvc.proto#L27). Each rpc request should have response. That because we have empty struct here.

pkg/mocks/drivemgr.go
```diff
@@ -48,18 +56,23 @@ func NewMockDriveMgrClient(drives []*api.Drive) *MockDriveMgrClient {
...
 // GetDrivesList returns provided to MockDriveMgrClient drives to imitate working of DriveManager
-func (m MockDriveMgrClient) GetDrivesList(ctx context.Context, in *api.DrivesRequest, opts ...grpc.CallOption) (*api.DrivesResponse, error) {
+func (m *MockDriveMgrClient) GetDrivesList(ctx context.Context, in *api.DrivesRequest, opts ...grpc.CallOption) (*api.DrivesResponse, error) {
 	return &api.DrivesResponse{
 		Disks: m.drives,
 	}, nil
 }
+
+// Locate is a stub for Locate DriveManager's method
+func (m MockDriveMgrClient) Locate(ctx context.Context, in *api.DriveLocateRequest, opts ...grpc.CallOption) (*api.DriveLocateResponse, error) {
```
painhardcore, line 76: You've made every method by a pointer and left this one by value. I think it's a good idea to make it in the same way

rususlasan: done

api/v1/drivemgrsvc.proto
```diff
@@ -13,6 +13,16 @@ message DrivesResponse {
     repeated Drive disks = 1;
 }
 
+message DriveLocateRequest {
+    string driveSerialNumber = 1;
+    string action = 2;
```
mishoyama, line 18: Using `string` here seems to be redundant. Suggest to have `integer` instead:

> 0 - start
> 1 - stop
> 2 - status

pkg/drivemgr/drivemgr_service.go
```diff
@@ -66,3 +65,14 @@ func (svc *DriveServiceServerImpl) GetDrivesList(ctx context.Context, req *api.D
...
+	err := svc.mgr.Locate(in.GetDriveSerialNumber(), in.GetAction())
```
mishoyama, line 71: I suggest to print waring when `codes.Unimplemented`

mishoyama (changes requested):

api/v1/drivemgrsvc.proto
rususlasan: done

mishoyama (approved):

pkg/drivemgr/drivemgr_service.go
rususlasan: I case of error we log message with ERROR level. Do you suggest to switch log level to WARNING if error codes is `unimplemented`? I think error level is suitable here.

painhardcore (approved):
LGTM

PR #203, opened 2020-12-10
[ISSUE-204] Add LVG support to DR procedure

ykulazhenkov:
## Purpose
Add support for drive replacement for LVG volumes

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [x] New unit tests added
- [ ] PR validation is passed
- [ ] Custom CI is passed

## Testing
_Link to custom CI build_

### DIFF
--- a/pkg/base/k8s/cr_helper.go
+++ b/pkg/base/k8s/cr_helper.go
@@ -26,6 +26,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 
 	api "github.com/dell/csi-baremetal/api/generated/v1"
+	apiV1 "github.com/dell/csi-baremetal/api/v1"
 	accrd "github.com/dell/csi-baremetal/api/v1/availablecapacitycrd"
 	"github.com/dell/csi-baremetal/api/v1/drivecrd"
 	"github.com/dell/csi-baremetal/api/v1/lvgcrd"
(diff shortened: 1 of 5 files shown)


mishoyama (approved):

pkg/base/k8s/cr_helper.go
```diff
@@ -105,29 +106,61 @@ func (cs *CRHelper) DeleteACsByNodeID(nodeID string) error {
...
+		return nil, err
 	}
-
+	var lvgUUID *string
 	for _, v := range volList.Items {
+		v := v
+		if v.Spec.LocationType == apiV1.LocationTypeLVM {
+			// lazy read
+			if lvgUUID == nil {
+				lvgUUID = cs.getLVGByDrive(ctx, location)
```
rususlasan, line 130: I can't understand that logic, why do you use `location` that was came as a parameter, as a location for each volume from list?

ykulazhenkov: Location is a drive ID. In this function we are always looking for a volumes which are located on that drive. If the volume is a "full-disk", we need to compare v.Spec.Location with location from function arg (drive ID). If volume is "LVM" based we can't use v.Spec.Location to filter volumes, because for such volumes LVG id stored in v.Spec.Location instead of drive id.  So, we need to find LVG which located on required drive and then filter volumes by the LVG id.
If we have no LVG based volumes we don't need to read LVG list at all, so we use lazy read here. We will read LVG list only once, when we find first lvm based volume (resolved lvgUUID will be placed to location var)

rususlasan: Still not clear for me, lets say we have next scenario:
- there are LVG-based on HDD-based volumes in the cluster, e.g. `vol1` - it is LVM - based volume, `vol2` - HDD
- function GetVolumesByLocation was called with `location` that points on HDD-based volume and we expect to get `vol2`, according for our example
- function read list of all volumes: `[vol1, vol2]`
- first loop iteration use `vol1`, as it was said before `vol1` is LVM-based 
- function run `cs.getLVGByDrive(ctx, location)` that return nil for current example
- and in line 132: `if lvgUUID == nil` function gets `True` and returns error

ykulazhenkov: You are right. Nice finding! 👍 Thank you! Fixed

Issue #209, opened 2020-12-16
Don't deploy scheduler extender in openshift_patcher.sh


mishoyama:
openshift_patcher.sh should patch scheduler config but not install extender to avoid issue with passing extra parameters

PR #210, opened 2020-12-16
[ISSUE-209] change openshift patcher logic

mishoyama:
### Issue #209 
OpenShift patcher shouldn't deploy scheduler extender but patch scheduler config only.

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [ ] New unit tests added
- [x] All TODOs are linked with the issues
- [x] All comments are resolved
- [x] PR validation is passed
- [ ] Custom CI is passed

## Testing
Manual deployment was done.


AnastasiaShemyakinskaya (approved):

docs/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,4 +1,7 @@
 ## Purpose
+<!--- Replace _Insert_Your_Issue_ID_ placeholder with your Issue i.e. ISSUE-1 -->
+### [_Insert_Your_Issue_ID_](https://github.com/dell/csi-baremetal/issues/_Insert_Your_Issue_ID_)
```
painhardcore, line 3: How this change is related to other changes?
However, making a full URL link to the issue isn't necessary since it could be achieved with a number sign i.e. #209
Also, it's a good idea to use [linking keywords](https://docs.github.com/en/free-pro-team@latest/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword)

painhardcore (changes requested):

docs/PULL_REQUEST_TEMPLATE.md
mishoyama: Not related. Just wanted to avoid copying URL each time.

ykulazhenkov:
We have `#!/bin/sh` shebang string, but using `function` keyword which is specific to `bash`. Can you include shebang fix to your PR? As an alternative, we can fix the code to be `sh` compatible. I prefer the second variant.

ykulazhenkov (commented):
We have `#!/bin/sh` shebang string, but using function keyword which is specific to bash. Can you include shebang fix to your PR? As an alternative, we can fix the code to be `sh` compatible. I prefer the second variant.

docs/PULL_REQUEST_TEMPLATE.md
mishoyama: @painhardcore please take a look again

mishoyama:
@ykulazhenkov lets do this under separate issue #214

painhardcore (approved):

PR #223, opened 2020-12-24
[ISSUE-224] change lvg detection method

rususlasan:
### Issue #224

Search all VGs on the system during discovery and if root fs doesn't locate on some VG set corresponding flag in volumemgr to false:
discoverLvgSSD = false

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [x] New unit tests added

## Testing
_Link to custom CI build_


pkg/node/volumemgr.go
```diff
@@ -886,56 +889,60 @@ func (m *VolumeManager) discoverLVGOnSystemDrive() error {
...
+	pvs, err := m.lvmOps.GetAllPVs()
 	if err != nil {
-		return fmt.Errorf(errTmpl, err)
+		return fmt.Errorf("unable to list PVs on the system: %v", err)
 	}
 
-	if devices[0].Rota != base.NonRotationalNum {
+	var systemPVName string
+	for _, pv := range pvs {
+		if strings.Contains(pv, driveCR.Spec.Path) {
```
ykulazhenkov, line 906: I think we should avoid using Contains here. If system has many drives we can hit error with collision between /dev/sda and /dev/sdaa

pkg/node/volumemgr.go
```diff
@@ -886,56 +889,60 @@ func (m *VolumeManager) discoverLVGOnSystemDrive() error {
...
-
+	// 3. check whether the system drive SSD or not
+	devices, err := m.listBlk.GetBlockDevices(driveCR.Spec.Path)
 	if err != nil {
-		return fmt.Errorf(errTmpl, err)
+		return fmt.Errorf("unable to get info about system drive: %s", err)
 	}
 
-	if !lvgExists {
+	if devices[0].Rota != base.NonRotationalNum {
```
ykulazhenkov, line 924: We already renamed our SC from system-ssdlvg to syslvg. Maybe we should remove this check and support system VG on all storage types?

rususlasan: good point, going to refactor...

pkg/node/volumemgr.go
rususlasan: Yes, I agreed, but we can't use `==` here because of PV might be created on some partition, e.g. `/dev/sda1`. I will handle case that you described.

rususlasan: I think we can rely on partition name like `DEVICE_NAME` + `NUBMER`
https://www.debian.org/releases/buster/amd64/apcs04.en.html#:~:text=Linux%20represents%20the%20primary%20partitions,drive%20is%20%2Fdev%2Fsda5%20.

rususlasan: refactored

pkg/node/volumemgr.go
rususlasan: refactored

pkg/node/volumemgr.go
ykulazhenkov: Maybe we can remove Contains check and use single regexp?
```regexp.Match(fmt.Sprintf("^%s\\d*$", a), driveCR.Spec.Path)```

rususlasan: yeah, done

ykulazhenkov (approved):

mishoyama (approved):

painhardcore (approved):

Issue #88, opened 2020-09-25
Handle disk health change for LVG volumes


rususlasan:
If disk's health that used in LVG change corresponding AC should be removed and Volumes health status should be changed.

PR #225, opened 2020-12-24
[ISSUE-106] Skip allocation for LVG in FAILED state by resetting AC size

mishoyama:
### Issue #106 
Skip allocation for LVG in FAILED state by resetting AC size
```
borism1@ubuntu:/workspace/csi-baremetal$ ac | grep 38a63540-48a3-4f08-a33f-7f3fde4c4382
fa9171c6-b72a-4009-b239-88abdb198dfd   84934656    HDDLVG         38a63540-48a3-4f08-a33f-7f3fde4c4382   2fd0ef69-d1a5-4faa-bf20-c8e4251de2de
/workspace/csi-baremetal$ lvgs
Id                                     Size        Status    Volumes                                                                                                                        Locations                                Node
38a63540-48a3-4f08-a33f-7f3fde4c4382   210763776   CREATED   [pvc-c3ec92df-18e5-419b-9daf-8419bae3a6f8 pvc-3fb2c138-b441-408e-b1f9-1e23e135b68b pvc-703e4c57-6409-442c-93b4-9e790dc4497c]   [691fd88b-f5b3-46ab-939c-83cace9189d7]   2fd0ef69-d1a5-4faa-bf20-c8e4251de2de

/workspace/csi-baremetal$ lvgs
Id                                     Size        Status   Volumes                                                                                                                        Locations                                Node
38a63540-48a3-4f08-a33f-7f3fde4c4382   210763776   FAILED   [pvc-c3ec92df-18e5-419b-9daf-8419bae3a6f8 pvc-3fb2c138-b441-408e-b1f9-1e23e135b68b pvc-703e4c57-6409-442c-93b4-9e790dc4497c]   [691fd88b-f5b3-46ab-939c-83cace9189d7]   2fd0ef69-d1a5-4faa-bf20-c8e4251de2de

/workspace/csi-baremetal$ ac | grep 38a63540-48a3-4f08-a33f-7f3fde4c4382
fa9171c6-b72a-4009-b239-88abdb198dfd   <none>      HDDLVG         38a63540-48a3-4f08-a33f-7f3fde4c4382   2fd0ef69-d1a5-4faa-bf20-c8e4251de2de
```

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [ ] New unit tests added
- [x] Modified code has meaningful comments
- [ ] All TODOs are linked with the issues
- [x] All comments are resolved

## Testing
_In progress_

PR #229, opened 2021-01-14
[ISSUE-88] Handle disk health change for lvg volumes

AnastasiaShemyakinskaya:
### Issue #88

* New field in LVG structure _Health_
* If drive health is changed, lvg health  and status are also changed

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [x] New unit tests added

### DIFF
--- a/pkg/crcontrollers/lvg/lvgcontroller.go
+++ b/pkg/crcontrollers/lvg/lvgcontroller.go
@@ -157,6 +157,7 @@ func (c *Controller) handlerLVGCreation(lvg *lvgcrd.LVG) (ctrl.Result, error) {
 	if locations, err = c.createSystemLVG(lvg); err != nil {
 		ll.Errorf("Unable to create system LVG: %v", err)
 		newStatus = apiV1.Failed
+		lvg.Spec.Health = apiV1.HealthBad
 	}
 	lvg.Spec.Status = newStatus
 	lvg.Spec.Locations = locations
--- a/pkg/crcontrollers/lvg/lvgcontroller.go
+++ b/pkg/crcontrollers/lvg/lvgcontroller.go
@@ -114,6 +114,10 @@ func (c *Controller) Reconcile(req ctrl.Request) (ctrl.Result, error) {
 		return ctrl.Result{}, c.resetACSizeOfLVG(lvg.Name)
 	}
 
+	if lvg.Spec.Health != apiV1.HealthGood {
+		return ctrl.Result{}, c.resetACSizeOfLVG(lvg.Name)
+	}
+
 	return ctrl.Result{}, nil
 }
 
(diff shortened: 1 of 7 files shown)


pkg/node/volumemgr.go
```diff
@@ -983,7 +983,25 @@ func (m *VolumeManager) handleDriveStatusChange(ctx context.Context, drive *api.
 			}
 		}
 	}
-
+	lvgs, err := m.crHelper.GetLVGCRs(m.nodeID)
+	if err != nil {
+		ll.Errorf("Failed get LVG CRs, error: %v", err)
+	} else {
+		for _, lvg := range lvgs {
+			if util.ContainsString(lvg.Spec.Locations, drive.UUID) {
```
ykulazhenkov, line 991: I think that we can stop loop after we find first LVG because, as far as I know, drive can belong to single LVG only. Also, we already have similar function in CR helper https://github.com/dell/csi-baremetal/blob/5642af2ef8a9d68ddad061227d4c4667fe097e22/pkg/base/k8s/cr_helper.go#L151 Maybe we can make the function public and use it here? Here we can leave only status update for LVG.

painhardcore (approved):

pkg/crcontrollers/lvg/lvgcontroller.go
```diff
@@ -157,6 +157,7 @@ func (c *Controller) handlerLVGCreation(lvg *lvgcrd.LVG) (ctrl.Result, error) {
 	if locations, err = c.createSystemLVG(lvg); err != nil {
 		ll.Errorf("Unable to create system LVG: %v", err)
 		newStatus = apiV1.Failed
+		lvg.Spec.Health = apiV1.HealthBad
```
mishoyama, line 160: This looks confusing. Do we use `Health` field for logical health as well? I also don't see where we ignore such LVGs for allocation...

mishoyama (changes requested):

pkg/crcontrollers/lvg/lvgcontroller.go
AnastasiaShemyakinskaya: We update LVG status to `Failed` in case of drive bad health. If LVG has `Failed` status. CSI resets its AC. That was implemented here https://github.com/dell/csi-baremetal/pull/225

pkg/node/volumemgr.go
AnastasiaShemyakinskaya: Done

ykulazhenkov (approved):

mishoyama (approved):

mishoyama:
@AnastasiaShemyakinskaya please go ahead and merge your changes.

Issue #222, opened 2020-12-23
Volume CR must be created in the same namespace as PVC/PV


AnastasiaShemyakinskaya:
```
borism1@ubuntu:/workspace/csi-baremetal$ pvc -n borism1
NAME         STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS       AGE
data-web-0   Bound    pvc-5ab52e71-38fd-4767-82a4-29a8eff0a0b3   8001Gi     RWO            baremetal-csi-sc   25s
logs-web-0   Bound    pvc-63cbf91e-49a2-45f8-962f-aa702f833b75   8001Gi     RWO            baremetal-csi-sc   25s
www-web-0    Bound    pvc-0508d617-fa33-4333-837d-6c08087e7cda   8001Gi     RWO            baremetal-csi-sc   25s
borism1@ubuntu:/workspace/csi-baremetal$ volumes
Id                                         Size            Type   Health   OpStatus    CSIStatus   Usage    Location                               Node
pvc-0508d617-fa33-4333-837d-6c08087e7cda   8591008333824   xfs    GOOD     OPERATIVE   PUBLISHED   IN_USE   977d465b-87a2-433b-bc76-9452ba3c6319   fb481c19-ff8d-45e7-b6bd-82be262fc6a6
pvc-5ab52e71-38fd-4767-82a4-29a8eff0a0b3   8591008333824   xfs    GOOD     OPERATIVE   PUBLISHED   IN_USE   26d8a0fe-21ce-4508-a5b3-7428b19a4605   fb481c19-ff8d-45e7-b6bd-82be262fc6a6
pvc-63cbf91e-49a2-45f8-962f-aa702f833b75   8591008333824   xfs    GOOD     OPERATIVE   PUBLISHED   IN_USE   532c2b45-64ce-45a1-8fbe-f16cbe881f44   fb481c19-ff8d-45e7-b6bd-82be262fc6a6
borism1@ubuntu:/workspace/csi-baremetal$ kubectl describe volume pvc-0508d617-fa33-4333-837d-6c08087e7cda
Name:         pvc-0508d617-fa33-4333-837d-6c08087e7cda
Namespace:    
Labels:       <none>
Annotations:  <none>
API Version:  baremetal-csi.dellemc.com/v1
Kind:         Volume
```
Otherwise POD might not be able query info about its volumes.

mishoyama:
@AnastasiaShemyakinskaya please add detailed description why we need these changes...

PR #234, opened 2021-01-15
[ISSUE-222] Volume cr must be created in the same namespace as pvc/pv

AnastasiaShemyakinskaya:
### Issue #222

_Describe your changes:_

* Make Volume namespaced in helm charts
* Recreate Volume in node stage request in case it doesn't have the same namespace as an according PVC
* In CreateVolume and DeleteVolume we search for Volume with List request in Kubernetes API, because we don't know its namespace on these steps.

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [x] New unit tests added


mishoyama (approved):

ykulazhenkov (approved):

pkg/base/cache/cache.go
```diff
@@ -0,0 +1,54 @@
...
+	c.items[key] = object
+}
+
+// Delete deletes object from MemCache with given key
+func (c *MemCache) Delete(key string) {
+	c.Lock()
+	defer c.Unlock()
+
+	delete(c.items, key)
+}
```
painhardcore, line 54: Please describe the new package and its purpose.
Also, cover it with unit tests.

painhardcore (changes requested):

pkg/base/cache/cache.go
AnastasiaShemyakinskaya: Done

painhardcore (approved):

Issue #269, opened 2021-02-10
Caching support in csi-driver


ykulazhenkov:
We should cache some information to decrease API call count.
This will helps us to achieve the required scalability and performance.

As a first step, we should analyze which CR can be cached.

- [x] Prepare proposal
- [x] Implement generic cached client
- [x] Add Volume, Dive, and AvailableCapacity CRs caching to Discovery loop in csi-node.
- [x] Add caching for PersistentVolumeClaim and Volume and StorageClass CRs in scheduler-extender
- [ ] Add POD and Node resources caching in csi-controller.

PR #270, opened 2021-02-10
[ISSUE-269] Caching proposal

ykulazhenkov:
### Issue https://github.com/dell/csi-baremetal/issues/269

_Describe your changes_

Initial version of caching proposal

### DIFF
--- a/docs/proposals/caching.md
+++ b/docs/proposals/caching.md
@@ -1,0 +1,78 @@
+# Proposal: Caching support
+
+Last updated: 10.02.2021
+
+## Proposal
+
+We should cache some information to decrease API call count.
+This will helps us to achieve required scalability and performance.
+
+### Kubernetes recourses used by CSI driver
+
+- AvailableCapacity (AC)
+- AvailableCapacityReservation (ACR)
+- Drive
+- LVG
+- Volume
+- Pod
+- Node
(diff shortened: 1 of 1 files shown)


AnastasiaShemyakinskaya:
Could we use Watch API with our CRs?

ykulazhenkov:
> Could we use Watch API with our CRs?

Yes, we can

AnastasiaShemyakinskaya:
> > Could we use Watch API with our CRs?
> 
> Yes, we can

Couldn't it help to prevent race in CSI component?

ykulazhenkov:
> Couldn't it help to prevent race in CSI component?
@AnastasiaShemyakinskaya 
Unfortunately, no. Driver components use these CRs for synchronization between each other.  Watch API allows us to update CR cache with low latency but can't provide warranties of reading the latest resource version from the cache. We can use Watch API for CR updates in some places, for example in "csi-node.Discover" loop but not in places where synchronization is required.

ykulazhenkov:
POC for CR caching in csi-node.Discover
https://github.com/dell/csi-baremetal/pull/274

AnastasiaShemyakinskaya (approved):

painhardcore (approved):

mishoyama (approved):

PR #274, opened 2021-02-11
[ISSUE-269] Use CR caching in csi-node Discover loop

ykulazhenkov:
### Issue https://github.com/dell/csi-baremetal/issues/269

Integrate cache CRHelper to volumemgr

All CRs read requests from csi-node Discover loop show receive cached data from the auto-updated cache

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [ ] New unit tests added


pkg/node/node.go
```diff
@@ -26,6 +26,7 @@ import (
 	"sync"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
+	csibmnodeconst "github.com/dell/csi-baremetal/pkg/crcontrollers/csibmnode/common"
```
AnastasiaShemyakinskaya, line 29: Wrong import format

cmd/node/main.go
```diff
@@ -115,8 +117,22 @@ func main() {
...
+	wrappedK8SClient := k8s.NewKubeClient(k8SClient, logger, *namespace)
+
+	k8sCache, err := k8s.GetK8SCache()
+	if err != nil {
+		logger.Fatalf("fail to create cache for kubernetes resources, error: %v", err)
+	}
+	// start cache
+	go func() {
+		// cache implementation we use newer returns err
+		_ = k8sCache.Start(stopCH)
```
mishoyama, line 129: If caching go routine died will we start using non-cached mode, read outdated information or just fail?

pkg/base/k8s/kubecache.go
```diff
@@ -0,0 +1,73 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	k8sCl "sigs.k8s.io/controller-runtime/pkg/client"
+	crApiutil "sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+)
+
+// GetK8SCache returns k8s cache
+func GetK8SCache() (cache.Cache, error) {
+	config := ctrl.GetConfigOrDie()
+	scheme, err := PrepareScheme()
```
mishoyama, line 33: Does it make sense to cache all CSI custom resources? Can we specify only particular?

pkg/base/k8s/kubecache.go
```diff
@@ -0,0 +1,73 @@
...
+}
+
+// KubeCache is a wrapper for controller-runtime cache
+type KubeCache struct {
+	k8sCl.Reader
+	log *logrus.Entry
+}
+
+// ReadCR CRReader implementation
+func (k KubeCache) ReadCR(ctx context.Context, name string, namespace string, obj runtime.Object) error {
```
mishoyama, line 56: How does caching work when we update CR though usual k8s client?

pkg/node/volumemgr.go
```diff
@@ -699,7 +706,7 @@ func (m *VolumeManager) discoverVolumeCRs() error {
 		"method": "discoverVolumeCRs",
 	})
 
-	driveCRs, err := m.crHelper.GetDriveCRs(m.nodeID)
+	driveCRs, err := m.cachedCrHelper.GetDriveCRs(m.nodeID)
```
mishoyama, line 709: Disk can be marked as OFFLINE in controller **statemonitor.go**. Is this really safe to read from cache here?

cmd/node/main.go
ykulazhenkov: Cache implements retry logic. If WATCH call to kube API fails, reconnection will happen automatically. If there is no connection at the moment old data still can be readable from the cache.

pkg/base/k8s/kubecache.go
ykulazhenkov: We don't cache all resources. It is just a schema preparation, so the client will know that these resources exist. Cache implementation will subscribe to resource updates only if this resource was read from the cache at least once.

pkg/node/volumemgr.go
ykulazhenkov: Cache is autoupdated, so we will receive this update.

pkg/base/k8s/kubecache.go
ykulazhenkov: Cache updates happen automatically in background. This is fully separate process from updates from another client instances.

mishoyama (approved):

pkg/node/node.go
ykulazhenkov: fixed

AnastasiaShemyakinskaya (approved):

painhardcore (approved):

PR #271, opened 2021-02-10
[ISSUE-268] Volume expansion proposal

AnastasiaShemyakinskaya:
### Issue #268 

Add proposal in docs for volume expansion

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [ ] New unit tests added
- [ ] Modified code has meaningful comments
- [ ] All TODOs are linked with the issues
- [ ] All comments are resolved

## Testing

Issue #268, opened 2021-02-09
Volume expansion POC and proposal


AnastasiaShemyakinskaya:
* POC for LVG volumes
* Docs page with proposal

PR #279, opened 2021-02-12
[ISSUE-296] Volume expansion implementation

AnastasiaShemyakinskaya:
### Issue #268
Implement volume expansion proposal https://github.com/dell/csi-baremetal/pull/271

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [ ] New unit tests added

## Testing
_Link to custom CI build_


pkg/base/linuxutils/lvm/lvm.go
```diff
@@ -116,6 +120,20 @@ func (l *LVM) PVRemove(name string) error {
...
 
+// ExpandLV expand logical volume
+// Receives full name of a logical volume and requiredSize to resize
+// Returns error if something went wrong
+func (l *LVM) ExpandLV(lvName string, requiredSize int64) error {
+	cmd := fmt.Sprintf(LVExpandCmdTmpl, strconv.FormatInt(requiredSize, 10), lvName)
+	_, stdErr, err := l.e.RunCmd(cmd,
+		command.UseMetrics(true),
+		command.CmdName(strings.TrimSpace(fmt.Sprintf(LVExpandCmdTmpl, "", ""))))
+	if err != nil && strings.Contains(stdErr, "No PV label found") {
```
ykulazhenkov, line 131: Could you please add a short comment in code about this case?

pkg/controller/controller.go
```diff
@@ -374,6 +376,66 @@ func (c *CSIControllerService) ListSnapshots(context.Context, *csi.ListSnapshots
...
+
+	if volume == nil {
+		return nil, status.Error(codes.NotFound, "Volume doesn't exist")
+	}
+
+	if !util.IsStorageClassLVG(volume.Spec.StorageClass) {
+		return nil, status.Error(codes.FailedPrecondition, "Volume doesn't have LVG storage class")
+	}
+
+	if volume.Spec.Size > req.GetCapacityRange().GetRequiredBytes() {
```
ykulazhenkov, line 408: Should we use [`AlignSizeByPE`](https://github.com/dell/csi-baremetal/blob/06d06aebc80e7116d06497db2fad96bb75eac175/pkg/base/capacityplanner/common.go#L33 ) for `req.GetCapacityRange().GetRequiredBytes()`?

pkg/controller/controller.go
```diff
@@ -374,6 +376,66 @@ func (c *CSIControllerService) ListSnapshots(context.Context, *csi.ListSnapshots
...
+	)
+
+	if volID == "" {
+		return nil, status.Error(codes.InvalidArgument, "Volume name missing in request")
+	}
+	if req.GetVolumeCapability() == nil {
+		return nil, status.Error(codes.InvalidArgument, "Volume capabilities missing in request")
+	}
+
+	volume := c.crHelper.GetVolumeByID(volID)
```
ykulazhenkov, line 398: I think we can try to improve error handling in all chHelper methods. Could you please check how hard it will be to change `GetVolumeByID` signature to return error alongside volume? I think it will be great if we can handle "not found" and "api call error" separately.

pkg/controller/controller.go
```diff
@@ -374,6 +376,66 @@ func (c *CSIControllerService) ListSnapshots(context.Context, *csi.ListSnapshots
...
+			CapacityBytes:         0,
+			NodeExpansionRequired: false,
+		}, nil
+	}
+
+	if err := c.k8sclient.ReadCR(ctxWithID, volume.Spec.Location, "", lvg); err != nil {
+		ll.Errorf("Failed to get LVG by location %s", volume.Spec.Location)
+		return nil, status.Error(codes.Internal, "Unable to read LVG")
+	}
+	if lvg.Spec.Size < req.GetCapacityRange().GetRequiredBytes() {
```
ykulazhenkov, line 423: Could you please describe why you are using lvg size here? Maybe it will be more correct to use LVG AC size here? Additionally, we need to decrease LVG AC size here.

pkg/controller/controller.go
```diff
@@ -374,6 +376,66 @@ func (c *CSIControllerService) ListSnapshots(context.Context, *csi.ListSnapshots
...
+	}
+	volume.Spec.CSIStatus = apiV1.Expanding
+	volume.Spec.Size = req.GetCapacityRange().GetRequiredBytes()
+
+	if err := c.k8sclient.UpdateCR(ctxWithID, volume); err != nil {
+		ll.Errorf("Failed to update LVG, error: %v", err)
+		return nil, status.Error(codes.Internal, "Unable to update volume")
+	}
+
+	if err := c.svc.WaitStatus(ctxWithID, volID, apiV1.Failed, apiV1.Expanded); err != nil {
```
ykulazhenkov, line 434: Will volume left with `Expanded` status after expand operation completes? Is it expected?

pkg/node/volumemgr.go
```diff
@@ -1213,3 +1215,38 @@ func (m *VolumeManager) addVolumeStatusAnnotation(drive *drivecrd.Drive, volumeN
...
+	if err = m.lvmOps.ExpandLV(volumePath, mbSize); err != nil {
+		volume.Spec.CSIStatus = apiV1.Failed
+		if updateErr := m.k8sClient.UpdateCRWithAttempts(ctx, volume, 10); updateErr != nil {
+			ll.Errorf("Unable to change volume %s status to %s, error: %v.",
+				volume.Name, volume.Spec.CSIStatus, updateErr)
+			return ctrl.Result{Requeue: true}, updateErr
+		}
+		return ctrl.Result{Requeue: true}, err
+	}
+	volume.Spec.CSIStatus = apiV1.Expanded
```
ykulazhenkov, line 1246: Could you please clarify why we need `Expanded` status? Do we need to know that volume resized after the expansion operation was fully completed? Can we use `created` as a generic status, which means "volume exists but not published"?

pkg/node/volumemgr.go
```diff
@@ -1213,3 +1215,38 @@ func (m *VolumeManager) addVolumeStatusAnnotation(drive *drivecrd.Drive, volumeN
...
+	if err != nil {
+		ll.Errorf("Failed to get volume path, err: %v", err)
+		return ctrl.Result{Requeue: true}, err
+	}
+	mbSize, err := util.ToSizeUnit(volume.Spec.Size, util.BYTE, util.MBYTE)
+	if err != nil {
+		ll.Errorf("Failed to convert %d to %v, err: %v", volume.Spec.Size, util.MBYTE, err)
+		return ctrl.Result{Requeue: true}, err
+	}
+	if err = m.lvmOps.ExpandLV(volumePath, mbSize); err != nil {
```
ykulazhenkov, line 1237: Is ExpandLV operation idempotent? If so, I think we can try to replace `UpdateCRWithAttempts` with `UpdateCR` function below and try to rely on  `ctrl.Result{Requeue: true}`.

pkg/node/volumemgr.go
AnastasiaShemyakinskaya: `ControllerExpandVolume` handles both online and offline volume expansion, so volume can be in published, created or volume_ready state. As I see we can't just set `created`, because the previous state of the volume could be published for example. I assume, we need to know if the volume was successfully expanded, so we can set its previous state after expansion.

pkg/common/volume_operations.go
```diff
@@ -424,6 +428,78 @@ func (vo *VolumeOperationsImpl) WaitStatus(ctx context.Context, volumeID string,
...
+		"method":   "ExpandVolume",
+		"volumeID": volume.Spec.Id,
+	})
+	currStatus := volume.Spec.CSIStatus
+	switch currStatus {
+	case apiV1.Resizing, apiV1.Resized:
+		ll.Debug("Volume is already expanding")
+	case apiV1.VolumeReady, apiV1.Created, apiV1.Published:
+		if !util.IsStorageClassLVG(volume.Spec.StorageClass) {
+			return nil, status.Error(codes.FailedPrecondition, "Volume doesn't have LVG storage class")
```
mishoyama, line 447: Please rename `Storage class <name> doesn't support volume expansion`. Does it make sense to check corresponding field in SC? User might create new LVG SC and disable expansion.

pkg/common/volume_operations.go
```diff
@@ -424,6 +428,78 @@ func (vo *VolumeOperationsImpl) WaitStatus(ctx context.Context, volumeID string,
...
+		}
+		capacity, err := vo.crHelper.GetACByLocation(volume.Spec.Location)
+		if err != nil {
+			ll.Errorf("Failed to get AC by location %s", volume.Spec.Location)
+			return nil, status.Error(codes.Internal, "Unable to read AC")
+		}
+
+		if capacity.Spec.Size < requiredBytes {
+			return nil, status.Error(codes.OutOfRange, "Required bytes are less than existing LVG size")
```
mishoyama, line 456: `Not enough capacity to expand volume: requested - %s, available - %s`

pkg/controller/controller.go
```diff
@@ -374,6 +376,57 @@ func (c *CSIControllerService) ListSnapshots(context.Context, *csi.ListSnapshots
 }
 
 // ControllerExpandVolume is not implemented yet
```
mishoyama, line 378: Please update method comments

pkg/common/volume_operations.go
```diff
@@ -424,6 +428,78 @@ func (vo *VolumeOperationsImpl) WaitStatus(ctx context.Context, volumeID string,
...
+		}
+		volume.Spec.CSIStatus = apiV1.Resizing
```
mishoyama, line 458: Shouldn't we reserve space in AC first and reject request right away if update AC failed?

mishoyama (changes requested):

pkg/common/volume_operations.go
AnastasiaShemyakinskaya: If SC doesn't support resizing, it doesn't have field `allowVolumeExpansion: true` in spec. In this case, csi-resizer doesn't call ControllerExpandVolume for unsupported SC. So this check is redundant because if SC allows volume expansion, it has a corresponding field in spec, and resizer call request only for such SC. But if User creates not LVG SC and enables expansion, it makes sense to check if volume SC is LVG.

pkg/common/volume_operations.go
AnastasiaShemyakinskaya: Yes, it makes sense

pkg/common/volume_operations.go
mishoyama: Ok, got it!

pkg/common/volume_operations.go
```diff
@@ -424,6 +428,78 @@ func (vo *VolumeOperationsImpl) WaitStatus(ctx context.Context, volumeID string,
...
+	ll := vo.log.WithFields(logrus.Fields{
+		"method":   "UpdateCRsAfterVolumeExpansion",
+		"volumeID": vol.Id,
+	})
+	namespace, err := vo.cache.Get(vol.Id)
+	if err != nil {
+		ll.Errorf("Failed to get namespace from cache, error: %v", err)
+		return err
+	}
+	if err = vo.crHelper.UpdateVolumeCRSpec(vol.Id, namespace, vol); err != nil {
```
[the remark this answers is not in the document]
AnastasiaShemyakinskaya: We update volume status and try to return its previous state. So after successful resizing volume won't be in `Resized` state but has its status before resizing. In Reconcile we can't determine its previous state

pkg/controller/controller.go
```diff
@@ -374,6 +376,57 @@ func (c *CSIControllerService) ListSnapshots(context.Context, *csi.ListSnapshots
...
+		}, nil
+	}
+
+	c.reqMu.Lock()
+	if volSpec, err = c.svc.ExpandVolume(ctx, volume, requiredBytes); err != nil {
+		return nil, err
+	}
+	c.reqMu.Unlock()
+
+	if err := c.svc.WaitStatus(ctxWithID, volID, apiV1.Failed, apiV1.Resized); err != nil {
```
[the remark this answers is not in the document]
AnastasiaShemyakinskaya: 1) We will try to return capacity in AC and Volume. I will implement it
2) In UpdateCRAfterVolumeExpansion it should be safe
3) If expansion was successful we return volume status to its previous value

pkg/common/volume_operations.go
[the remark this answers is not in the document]
mishoyama: Do you store previous state in cache? How do you guarantee that operation will be done until context is expired. What will happen if controller restarts? If really required you can store previous state in annotations.

pkg/common/volume_operations.go
```diff
@@ -424,6 +428,78 @@ func (vo *VolumeOperationsImpl) WaitStatus(ctx context.Context, volumeID string,
...
+		ll.Errorf("Failed to update volume, error: %v", err)
+		return err
+	}
+	capacity, err := vo.crHelper.GetACByLocation(vol.Location)
+	if err != nil {
+		ll.Errorf("Failed to get AC by location %s, error: %v", vol.Location, err)
+		return err
+	}
+	capacity.Spec.Size -= vol.Size
```
mishoyama, line 495: Do you align size of expansion to PE size?

AnastasiaShemyakinskaya: It's done here https://github.com/dell/csi-baremetal/blob/739d02e009c5d2c8cdf0525e5fd849a1a3bd5444/pkg/controller/controller.go#L388

pkg/common/volume_operations.go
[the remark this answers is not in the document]
AnastasiaShemyakinskaya: Ok, will try to use annotations

mishoyama (approved):

safronovD (approved):

Issue #280, opened 2021-02-15
LVM metadata size ignored when drive AvailableCapacity converted to LVM AvailableCapacity


ykulazhenkov:
LVM metadata size ignored when drive AvailableCapacity converted to LVM AC

We should decrease created AC size on LVM metadata size.

https://github.com/dell/csi-baremetal/blob/922f0a765abbf649cf3f1aebe4937e2068195619/pkg/common/ac_operations.go#L74

```
# CSI CRs
$ csi-lvg 
Name                                   Size        Status    Health   Locations                           
7114e5f3-671d-4b74-893a-50faa3049ac2   524288000   CREATED   GOOD     [d491c7bb-447c-4cc5-931c-08b2cf52bc75]   

$ csi-ac 
Name                                   Size        StorageClass    Location
1d262c3a-474a-45c1-ba61-5739193eb045   104857600   HDDLVG          7114e5f3-671d-4b74-893a-50faa3049ac2
```

```
# state on node
root@localdev:~# vgs --units B
  VG                                   #PV #LV #SN Attr   VSize        VFree     
  7114e5f3-671d-4b74-893a-50faa3049ac2   1   4   0 wz--n-   520093696B 100663296B
```

PR #281, opened 2021-02-16
[ISSUE-280] Fix AvailableCapacity size computation for LVM

ykulazhenkov:
### Issue #280 

Add SubtractLVMMetadataSize function, which allows to correctly calculate capacity which will be available after LVM creation (when default Metadata and PE size used)

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [x] New unit tests added

### DIFF
--- a/pkg/base/capacityplanner/common.go
+++ b/pkg/base/capacityplanner/common.go
@@ -38,3 +38,13 @@ func AlignSizeByPE(size int64) int64 {
 	}
 	return size + alignement
 }
+
+// SubtractLVMMetadataSize subtracts LVM metadata size from raw drive size
+func SubtractLVMMetadataSize(size int64) int64 {
+	reminder := size % DefaultPESize
+	result := size - reminder
+	if reminder < LvgDefaultMetadataSize {
+		return result - DefaultPESize
+	}
+	return result
+}
--- a/pkg/base/capacityplanner/node_capacity.go
+++ b/pkg/base/capacityplanner/node_capacity.go
@@ -107,17 +107,15 @@ func (nc *nodeCapacity) selectACForVolume(vol *genV1.Volume) *accrd.AvailableCap
 	// try to find free capacity
 	var foundAC *accrd.AvailableCapacity
 	for _, acs := range filteredMap {
-		foundAC = searchACWithClosestSize(acs, size)
+		foundAC = searchACWithClosestSize(acs, size, nil)
 		if foundAC != nil {
 			break
 		}
 	}
 	// for LVG SC try to reserve AC to create new LVG since no free space found on existing
 	if isLVG && foundAC == nil {
-		// for the new lvg we need some extra space
-		size += LvgDefaultMetadataSize
 		// search AC in sub storage class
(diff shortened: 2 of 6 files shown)


ykulazhenkov:
```
# VG on node
root@localdev:~# vgs --units b
  VG                                   #PV #LV #SN Attr   VSize        VFree     
  1d1fb5de-9fc5-41ed-ae4e-47f47160f4ba   1   4   0 wz--n-   520093696B 100663296B
```

```
# csi CRs
$ csi-ac
Name                                   Size        StorageClass     Location
86a1a6f0-2d4d-4415-b5a1-da4056f3ed2f   100663296   HDDLVG           1d1fb5de-9fc5-41ed-ae4e-47f47160f4ba
$ csi-lvg 
Name                                   Size        Status    Health      
1d1fb5de-9fc5-41ed-ae4e-47f47160f4ba   520093696   CREATED   GOOD  
```

pkg/base/capacityplanner/common.go
```diff
@@ -38,3 +38,13 @@ func AlignSizeByPE(size int64) int64 {
 	}
 	return size + alignement
 }
+
+// SubtractLVMMetadataSize subtracts LVM metadata size from raw drive size
+func SubtractLVMMetadataSize(size int64) int64 {
+	reminder := size % DefaultPESize
+	result := size - reminder
+	if reminder < LvgDefaultMetadataSize {
+		return result - DefaultPESize
```
mishoyama, line 47: Why do we increase by PE size and not LVG metadata size?

pkg/base/capacityplanner/node_capacity.go
```diff
@@ -107,17 +107,15 @@ func (nc *nodeCapacity) selectACForVolume(vol *genV1.Volume) *accrd.AvailableCap
...
 			break
 		}
 	}
 	// for LVG SC try to reserve AC to create new LVG since no free space found on existing
 	if isLVG && foundAC == nil {
-		// for the new lvg we need some extra space
-		size += LvgDefaultMetadataSize
 		// search AC in sub storage class
-		foundAC = searchACWithClosestSize(scToACMap[subSC], size)
+		foundAC = searchACWithClosestSize(scToACMap[subSC], size, SubtractLVMMetadataSize)
```
mishoyama, line 118: It's really hard to understand the logic in `selectACForVolume`. Does it make sense to refactor and split code for LVG and non LVG case?

mishoyama (approved):

pkg/base/capacityplanner/common.go
ykulazhenkov: We should round down AC size to PE extent size after we subtract metadata size.

pkg/base/capacityplanner/node_capacity.go
ykulazhenkov: I agree function is quite complex. I think refactoring can be nontrivial, so it will be better to do this as a separate activity.

AnastasiaShemyakinskaya (approved):

painhardcore (approved):

Issue #282, opened 2021-02-17
Add support for new lsblk version


mishoyama:
When building node image on top of Ubuntu 20.04 faced an issue with unmarshalling `lsblk` output:
```
Feb  1 13:07:36.626 [ERRO] Discover finished with error: discoverVolumeCRs return error: unable to inspect system block devices via lsblk, error: unable to unmarshal output to BlockDevice instance, error: json: cannot unmarshal number into Go struct field BlockDevice.children.size of type string
```
Latest lbslk version has slightly different output format:
```
lsblk --version
lsblk from util-linux 2.34

root@baremetal-csi-node-dwmff:/# lsblk /dev/sdc --paths --json --bytes --fs --output NAME,TYPE,SIZE,ROTA,SERIAL,WWN,VENDOR,MODEL,REV,MOUNTPOINT,FSTYPE,PARTUUID
{
   "blockdevices": [
      {"name":"/dev/sdc", "type":"disk", "size":8001563222016, "rota":true, "serial":"5000cca0bbce17ff", "wwn":"0x5000cca0bbce17ff", "vendor":"ATA     ", "model":"HGST_HUS728T8TAL", "rev":"RT04", "mountpoint":null, "fstype":null, "partuuid":null}
   ]
}

root@baremetal-csi-node-hklnz:/# lsblk --version
lsblk from util-linux 2.31.1

lsblk /dev/sdc --paths --json --bytes --fs --output NAME,TYPE,SIZE,ROTA,SERIAL,WWN,VENDOR,MODEL,REV,MOUNTPOINT,FSTYPE,PARTUUID
{
   "blockdevices": [
      {"name": "/dev/sdc", "type": "disk", "size": "8001563222016", "rota": "1", "serial": "5000cca0bbce87e7", "wwn": "0x5000cca0bbce87e7", "vendor": "ATA     ", "model": "HGST HUS728T8TAL", "rev": "RT04", "mountpoint": null, "fstype": null, "partuuid": null}
   ]
}
```

PR #283, opened 2021-02-17
[ISSUE-282] new lsblk version support

mishoyama:
### Issue #282

Add support for `lsblk` version 2.34

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [x] New unit tests added

## Testing
In progress


ykulazhenkov:
What do you think about using `interface{}` for fields with unknown types and then do a type assertion to convert them to the type required by the driver?
Example:
```
type dataInternal struct {
	Foo interface{} `json:"foo,omitempty"`
	Bar bool  `json:"bar,omitempty"`
}

type Data struct {
	Foo int
	Bar bool
}

func  (di *dataInternal) Data() (Data, error) {
	d := Data{Bar: di.Bar}
	switch v := di.Foo.(type) {
	case float64:
		d.Foo = int(v)
	case string:
		foo, err := strconv.Atoi(v)
		if err != nil {
			return Data{}, err
		}
		d.Foo = foo
	default:
		return Data{}, errors.New("unknown type for Foo")
	}
	return d, nil
}
```
This approach can remove the need to do a check of lsblk version.

mishoyama:
@ykulazhenkov looks good until output format of `lsblk` change significantly. Will try to modify code according to your proposal.

Issue #312, opened 2021-03-05
Add deployment based on OS kernel version


AnastasiaShemyakinskaya:
* Read node OS kernel labels
* Deploy corresponding daemon set

PR #315, opened 2021-03-10
[ISSUE-312] Add deployment based on os kernel version

AnastasiaShemyakinskaya:
### Issue #312 

_Describe your changes_

1) Add `CSIInstaller` structure and `Observer` interface
2) `CSIInstaller` implements `Observer` interface
3) `CSIInstaller` contains functions:
* `Notify` receives kernel version and sends it to `updated` channel
* `Install` calls install function in go routine and waits until it returns error value
* `install` waits until  `updated` channel received version or until context is done
* `intallWithHelm` installs CSI using Helm cmd
4) BM controller calls `Notify` when it gets kernel version for node
5) `Install` function is called in operator main 

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [x] New unit tests added

## Testing
_Link to custom CI build_

### DIFF
--- a/pkg/common/installer.go
+++ b/pkg/common/installer.go
@@ -83,8 +83,7 @@ func (c *CSIInstaller) install(ctx context.Context) error {
 		ll.Infof("Receive kernel version: %s", version)
 		return c.installWithHelm(version)
 	case <-ctx.Done():
-		ll.Info("Context is done, install without kernel version")
-		return c.installWithHelm("")
+		return fmt.Errorf("context is done: %v", ctx.Err())
 	}
 }
 
@@ -92,10 +91,7 @@ func (c *CSIInstaller) install(ctx context.Context) error {
 // Receive string
 // Return error
 func (c *CSIInstaller) installWithHelm(kernelVersion string) error {
-	cmd := fmt.Sprintf(HelmInstallCSICmdTmpl, c.version, c.drivemgr)
-	if kernelVersion != "" {
-		cmd = fmt.Sprintf(HelmInstallCSICmdTmpl, c.version, c.drivemgr) + fmt.Sprintf(KernelValue, kernelVersion)
-	}
+	cmd := fmt.Sprintf(HelmInstallCSICmdTmpl, c.version, c.drivemgr) + fmt.Sprintf(KernelValue, kernelVersion)
 	executor := command.NewExecutor(c.log.Logger)
 	if _, _, err := executor.RunCmd(cmd); err != nil {
 		return err
--- a/pkg/common/installer.go
+++ b/pkg/common/installer.go
@@ -94,7 +94,7 @@ func (c *CSIInstaller) install(ctx context.Context) error {
 	}
 }
 
-// installWithHelm tris to install helm with kernel version or without it if parameter kernelVersion is empty
+// installWithHelm tries to install helm with kernel version
 // Receive string
 // Return error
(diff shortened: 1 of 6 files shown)


pkg/crcontrollers/operator/controller.go
```diff
@@ -455,6 +457,9 @@ func (bmc *Controller) updateNodeLabelsAndAnnotation(k8sNode *coreV1.Node, goalV
 			k8sNode.Labels[common.NodeKernelVersionLabelKey] = version
 			toUpdate = true
 		}
+		if bmc.observer != nil {
+			bmc.observer.Notify(version)
```
mishoyama, line 461: Do we really need to notify each time? Does it make sense to do this only when kernel label isn't set or changed?

pkg/common/installer.go
```diff
@@ -0,0 +1,104 @@
...
+		return c.installWithHelm("")
+	}
+}
+
+// installWithHelm tris to install helm with kernel version or without it if parameter kernelVersion is empty
+// Receive string
+// Return error
+func (c *CSIInstaller) installWithHelm(kernelVersion string) error {
+	cmd := fmt.Sprintf(HelmInstallCSICmdTmpl, c.version, c.drivemgr)
+	if kernelVersion != "" {
```
mishoyama, line 96: Kernel version will never be empty here. And you can't jus pass this value since we have `default and `5.4` images only. Here we need to have some mapping.

mishoyama (changes requested):

pkg/common/installer.go
```diff
@@ -0,0 +1,104 @@
...
+	ll := c.log.WithFields(logrus.Fields{
+		"method": "install",
+	})
+	select {
+	case version := <-c.updated:
+		ll.Infof("Receive kernel version: %s", version)
+		return c.installWithHelm(version)
+	case <-ctx.Done():
+		ll.Info("Context is done, install without kernel version")
+		return c.installWithHelm("")
```
mishoyama, line 87: Why do we need to install w/o version when context is done?

pkg/common/installer.go
```diff
@@ -0,0 +1,104 @@
...
+	}
+}
+
+// installWithHelm tris to install helm with kernel version or without it if parameter kernelVersion is empty
+// Receive string
+// Return error
+func (c *CSIInstaller) installWithHelm(kernelVersion string) error {
```
mishoyama, line 94: When all nodes have the same kernel version you will run `helm install` multiple times. We need to check first before running command and protect against race conditions...

mishoyama (changes requested):

pkg/common/installer.go
AnastasiaShemyakinskaya: We call `installWithHelm` if we get something in `updated` channel. This channel received data in `Notify` function. In Notify we use  by Do() function from sync.Once. This function is called only once in multithreaded environment. So `helm install` will be called once, when it gets kernel version from controller.

pkg/common/installer.go
AnastasiaShemyakinskaya: I supposed if something goes wrong with labels setting in nodes and we can't get kernel version, we install CSI with default settings. Should we return error instead?

mishoyama: Yes, we need to return error in this case.

mishoyama (approved):

safronovD (approved):

Issue #326, opened 2021-03-19
Add debug logging in PrepareAndPerformMount


safronovD:
Improve logs to find a bug while executing `mount` in `NodePublishVolume` stage

PR #327, opened 2021-03-19
[ISSUE-326] Add debug logging in PrepareAndPerformMount

safronovD:
### Issue #326 

Add debug logs if mount operation fails. Check:
- if src exists
- if src is mounted
- mounted device (to src) has space

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [ ] New unit tests added

## Testing
_Link to custom CI build_


pkg/node/provisioners/utilwrappers/fs_operations.go
```diff
@@ -96,8 +96,37 @@ func (fsOp *FSOperationsImpl) PrepareAndPerformMount(src, dst string, bindMount,
 	}
 	if err := fsOp.Mount(src, dst, opts); err != nil {
 		if wasCreated {
-			_ = fsOp.RmDir(dst)
+			err = fsOp.RmDir(dst)
```
AnastasiaShemyakinskaya, line 99: Why do we need err here, if we don't check it or use it?

safronovD: fixed

pkg/node/provisioners/utilwrappers/fs_operations.go
```diff
@@ -98,6 +98,35 @@ func (fsOp *FSOperationsImpl) PrepareAndPerformMount(src, dst string, bindMount,
...
+
+		isSrcMounted, err := fsOp.IsMounted(src)
+		if err != nil {
+			ll.Debugf("failed to execute isMount: %s", err)
+		}
+		if !isSrcMounted {
+			ll.Debugf("Src is not mounted")
+		} else {
+			if srcMount, err := fsOp.FindMountPoint(src); err != nil {
+				ll.Debugf("failed to find mountPoint for src: %s", err)
```
mishoyama, line 119: Are those `failed` log messages errors? Do we need to print warnings at least?

safronovD: I thought, all this block would print optional debug info. But with warnings it looks better, it's not possible to fail these operations in normal state. Fixed

mishoyama (approved):

AnastasiaShemyakinskaya (approved):

PR #465, opened 2021-07-08
[ISSUE-464] Scheduler-extender readiness check

safronovD:
### Issue #464 

Implement proposal of implementing scheduler-extender readiness check (based on kube-scheduler restart):
- add readiness probe to extender
- patcher try to recreate kube-scheduler manifest if not changed
- remove kube-scheduler restart check from e2e

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [x] New unit tests added
- [x] Modified code has meaningful comments
- [ ] All TODOs are linked with the issues
- [ ] All comments are resolved

## Testing
_Link to custom CI build_

Issue #464, opened 2021-07-08
Implement scheduler-extender readiness check


safronovD:
Error in CSI e2e tests:
```
/var/jenkins_home/workspace/csi-master-ci/test/e2e/scenarios/stress-test.go:109
Unexpected error:
    <*errors.errorString | 0xc0003ec9f0>: {
        s: "scheduler is not restarted",
    }
    scheduler is not restarted
occurred
/var/jenkins_home/workspace/csi-master-ci/test/e2e/scenarios/bare-metal-testdriver.go:97
```

Scheduler restart check in test may not be correct sometimes. It's necessary to move this logic to Operator or extender readiness check.

PR #478, opened 2021-08-04
[ISSUE-464] Retry patching after kube-scheduler not restarted

safronovD:
Signed-off-by: safronovD <danil_safronov@dell.com>

### Issue #464 

RCA:
After patching process we had the race, when kube-scheduler restarted in 8+ minutes (due to many configuration changes while csi installing/deleting). In this case kube-scheduler stacked in the patched setup from previous run and updating/recreating pod manifest wasn't affected it.
To avoid this situation new functional of repeat patching after kube-scheduler inactivity was supported in https://github.com/dell/csi-baremetal-operator/pull/47 and https://github.com/dell/csi-baremetal/pull/465. Also, we need to use this option in e2e to fix the problem

Changes:
- Set Patching retry timeout to 3 mins
- Increase deployment wait timeout to 6 mins

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [ ] New unit tests added

## Testing
_Link to custom CI build_

### DIFF
--- a/test/e2e/common/deploy.go
+++ b/test/e2e/common/deploy.go
@@ -105,14 +105,16 @@ func DeployOperator(f *framework.Framework) (func(), error) {
 // DeployCSI deploys csi-baremetal-deployment with CmdHelmExecutor
 // After install - waiting all pods ready, checking kubernetes-scheduler restart
 // Cleanup - deleting csi-chart, cleaning all csi custom resources
-// Helm command - "helm install csi-baremetal <CHARTS_DIR>/csi-baremetal-deployment
+// Helm command - helm install csi-baremetal <CHARTS_DIR>/csi-baremetal-deployment
 // 			--set image.tag=<CSI_VERSION>
-//			--set image.pullPolicy=IfNotPresent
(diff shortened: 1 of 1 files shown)


mishoyama:
Please add some RCA details. Why do we need to change timeouts?

yurkov-anton (approved):

test/e2e/common/deploy.go
```diff
@@ -123,13 +123,15 @@ func DeployCSI(f *framework.Framework, additionalInstallArgs string) (func(), er
 			path:      path.Join(BMDriverTestContext.ChartsDir, "csi-baremetal-deployment"),
 			namespace: f.Namespace.Name,
 		}
-		installArgs = fmt.Sprintf("--set image.tag=%s "+
+		seReadinessTimeout = 3 // Minutes
```
mishoyama, line 126: Please add comments.

mishoyama (approved):

test/e2e/common/deploy.go
safronovD: updated

PR #565, opened 2021-10-06
[ATL-2199] Release workflow

misanthropicat:
### Issue #

_Describe your changes_

## PR checklist
- [ ] Add link to the issue
- [ ] Choose PR label
- [ ] New unit tests added

## Testing
_Link to custom CI build_

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -27,7 +27,6 @@ jobs:
         id: vars
         run: |
           echo "::set-output name=version::$(make version)"
-          echo "::set-output name=commit_hash::$(git rev-parse --short HEAD)"
 
       - name: Create asset
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,62 @@
...
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+      
+      - name: Make version
+        id: vars
+        run: |
+          echo "::set-output name=version::$(make version)"
+          echo "::set-output name=commit_hash::$(git rev-parse --short HEAD)"
```
rkiselenko, line 30: Can we use `github.sha` here?

.github/workflows/release.yml
```diff
@@ -0,0 +1,62 @@
...
+
+      - name: Create asset
+        run: |
+          PACKAGE_NAME=csi-baremetal-deployment-${{ github.event.inputs.csi_operator_version }}.tgz
+          DIR=/tmp/csi-baremetal-deployment-${{ steps.vars.outputs.version }}
+          cp -r ./charts $DIR
```
rkiselenko, line 36: It's better to use `artifacts` or `cache` actions here. There is no guarantee that files in `/tmp` folder will be persist between jobs.

.github/workflows/release.yml
```diff
@@ -0,0 +1,61 @@
+name: Release Charts
```
mishoyama, line 1: @misanthropicat please use GitHub issues instead if Jira

Issue #567, opened 2021-10-07
Create release workflow for charts


misanthropicat:
### Issue title

Automate CSI charts release procedure.

### Environment overview

*Please, include the information available about your environment using the format below:*

* K8S version:
* Node OS: 
* Helm: 
* K8S cluster overview (# of nodes, resources available):
* Other relevant software installed (with the software versions):
* Any other constraints we need to know: 

### Issue statement

* Introduction (*What is the solution you're developing? What is the aim you're trying to achieve?*): 
* Actions performed:
* Desired system behavior: 
* Observed system behavior:
* Research that you've preliminary done: 

### Code and data

*Ideally the code you're sharing should:*
* Be standalone (no dependencies or all the dependencies are included);
* Be complete (no missing imports or directives);
* Actually compile;
* Demonstrate the problem.

### Other useful information

Please, refer to [Writing the perfect question](https://codeblog.jonskeet.uk/2010/08/29/writing-the-perfect-question/) for more inspiration

PR #568, opened 2021-10-07
[ISSUE-567] Release workflow

misanthropicat:
### Issue #567 

Created manually (or by GitHub API) triggered workflow for charts releasing

## PR checklist
- [x] Add link to the issue
- [x] Choose PR label
- [ ] New unit tests added

## Testing
[successful release workflow run on fork](https://github.com/misanthropicat/csi-baremetal/actions/runs/1358541603) triggered [this workflow in operator repo](https://github.com/misanthropicat/csi-baremetal-operator/actions/runs/1358542842). [Release in operator repo](https://github.com/misanthropicat/csi-baremetal-operator/releases/tag/v0.5.0) (no changes in changelog due to no pull requests in fork repo, but in the action log correct tags appeared, so it should work in base repo properly)
image


.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
+      
+      - name: Make version
+        id: vars
+        run: |
+          echo "::set-output name=version::$(make version)"
+
+      - name: Create asset
+        run: |
+          PACKAGE_NAME=csi-baremetal-deployment-${{ github.event.inputs.csi_operator_version }}.tgz
```
mishoyama, line 33: Where is the second asset?

.github/workflows/release.yml
```diff
@@ -0,0 +1,68 @@
...
+          DIR=./csi-baremetal-deployment-${{ steps.vars.outputs.version }}
+          cp -r ./charts $DIR
+          tar -czvf ./$PACKAGE_NAME $DIR
```
mishoyama, line 36: To make package you should use helm command

.github/workflows/release.yml
misanthropicat: It is gathered in operator repo by release.yml workflow which is currently in progress.

.github/workflows/release.yml
misanthropicat: Ok, and what about version? Do I need to use only operator version for `helm package --version=${{ github.event.inputs.csi_operator_version }}`?

.github/workflows/release.yml
```diff
@@ -0,0 +1,53 @@
...
+
+jobs:
+  build:
+    name: Release
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+
+      - name: Create issue
```
mishoyama, line 16: I think we can avoid issue creation

rkiselenko (approved):
LGTM

PR #640, opened 2021-12-07
[ISSUE-637] Support k8s 1.22 version for csi

### LINKED ISSUE #637: Support CSI Baremetal on k8s 1.22
Currently CSI Baremetal not works on k8s 1.22. Need to support k8s 1.22 at CSI.

Need to update sidecar images.

**Additional context**

Aber4Nod:
Signed-off-by: Nikita Mikhnenko <nikita_mikhnenko@dell.com>

### Resolves #637 

Updating csi-provisioner version to v3.0.0 to support k8s v1.22.0

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [ ] Choose PR label
- [ ] New unit tests added

## Testing
_Provide test details_

### DIFF
--- a/variables.mk
+++ b/variables.mk
@@ -22,7 +22,7 @@ TAG              := ${FULL_VERSION}
 BRANCH           := $(shell git rev-parse --abbrev-ref HEAD)
 
 ### third-party components version
-CSI_PROVISIONER_TAG := v1.6.0
+CSI_PROVISIONER_TAG := v3.0.0
 CSI_RESIZER_TAG     := v1.1.0
 CSI_REGISTRAR_TAG   := v1.3.0
 LIVENESS_PROBE_TAG  := v2.1.0


variables.mk
```diff
@@ -22,7 +22,7 @@ TAG              := ${FULL_VERSION}
 BRANCH           := $(shell git rev-parse --abbrev-ref HEAD)
 
 ### third-party components version
-CSI_PROVISIONER_TAG := v1.6.0
```
mishoyama, line 25: Old tag is still in use, right? https://github.com/dell/csi-baremetal-operator/blob/21a6f06652587b1e426ff9094f3f6b1d56882bd3/charts/csi-baremetal-deployment/templates/csi-baremetal_v1_deployment.yaml#L45

Aber4Nod: Yes, but for out kind version 1.18 it's ok to use provisioner version 3.0.

mishoyama: Do you mean the provisioner 3.0 will work with k8s 1.18 as well?

Aber4Nod: Yes, it can. v1 api version of CSINode crd was introduced in k8s 1.17

mishoyama: In this case why do we need to have k8s version check here - https://github.com/dell/csi-baremetal-operator/blob/master/charts/csi-baremetal-deployment/templates/csi-baremetal_v1_deployment.yaml#L42 and deprecate tag here - https://github.com/dell/csi-baremetal-operator/blob/master/charts/csi-baremetal-deployment/values.yaml#L83 if minimal supported version is 1.18?

Aber4Nod: If we do not support the earlier versions - than there is no need in it

mishoyama (approved):

safronovD (approved):

Issue #673, opened 2021-12-23
Trigger e2e test on merge to master


rkiselenko:
Since e2e trigger from comment with body `/ci`, need run e2e on merge to master.

**Expected behavior**
Merge to master wont fail e2e due to empty comment

**Screenshots**
Not applicable

**Additional context**
Nope

PR #674, opened 2021-12-23
[ISSUE-647] e2e test artifacts

### LINKED ISSUE #647: Attach new e2e test logs to GitHub action artifacts
Make artifacts available in GH workflow

rkiselenko:
### Resolves #647 #673

Make artifacts available in GH workflow
Fix on comment trigger git ref
Change README badge

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [x] Choose PR label
- [ ] New unit tests added

## Testing
_Provide test details_

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -30,7 +30,7 @@ jobs:
         result-encoding: string
         script: |
           const { COMMENT_BODY, KEY } = process.env
-          var branch_line = `${COMMENT_BODY}`.split(String.fromCharCode(0x0A))[1]
+          var branch = `${COMMENT_BODY}`.split(String.fromCharCode(0x0A))[1]
           var key = `${KEY}`
           if (branch){
             if (branch.search(key) !== -1){
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -183,6 +183,7 @@ jobs:
 
     - name: Publish Unit Test Results
       uses: EnricoMi/publish-unit-test-result-action@v1
+      if: always()
       with:
         files: test/e2e/reports/report.xml
 
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -1,9 +1,6 @@
 name: e2e
 
 on:
-  pull_request:
-    branches:
-      - master
   push:
     branches: [ master ]
   issue_comment:
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -183,7 +183,6 @@ jobs:
 
     - name: Publish Unit Test Results
       uses: EnricoMi/publish-unit-test-result-action@v1
-      if: always()
       with:
         files: test/e2e/reports/report.xml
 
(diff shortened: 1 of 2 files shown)


mishoyama (changes requested):
Please update description in https://github.com/dell/csi-baremetal/issues/673 according to template

.github/workflows/e2e.yml
```diff
@@ -40,6 +40,7 @@ jobs:
       # Since the issue comment event runs against the main branch
       # we need to figure out a way to checkout the correct revision 
       # if we want to work with the changes from a pull request.
+      if: github.event.issue.pull_request && startsWith(github.event.comment.body, '/ci')
```
mishoyama, line 43: Is it possible to perform this check once and set global variable?

rkiselenko:
> Please update description in #673 according to template

Done

.github/workflows/e2e.yml
rkiselenko: That's check is matter only in context of step

mishoyama: It doesn't matter - we need to avoid code duplication when possible. Workflow is started in 2 cases:
1. Push in master branch
2. Comment `/ci` in PR
Two conditions can be checked at the beginning of workflow

rkiselenko: > `if` to prevent a job from running unless a condition is met. 

All `if` in steps handles `push` and `comment` events.

What is exactly do you suggest?

yurkov-anton (approved):

safronovD (approved):

docs/README.md
```diff
@@ -1,5 +1,5 @@
 [![PR validation](https://github.com/dell/csi-baremetal/actions/workflows/pr.yml/badge.svg)](https://github.com/dell/csi-baremetal/actions/workflows/pr.yml)
-[![E2E testing](https://github.com/dell/csi-baremetal/actions/workflows/e2e.yml/badge.svg)](https://github.com/dell/csi-baremetal/actions/workflows/e2e.yml)
+[![E2E testing](https://github.com/dell/csi-baremetal/actions/workflows/e2e.yml/badge.svg?branch=master&event=push)](https://github.com/dell/csi-baremetal/actions/workflows/e2e.yml)
```
mishoyama, line 2: How will this work for release branches?

.github/workflows/e2e.yml
mishoyama: I suggest to use variable like `is_pr_validation` and set it to true at the beginning of workflow when `github.event.issue.pull_request && startsWith(github.event.comment.body, '/ci')` is true and use this variable further instead of repeating `if: github.event.issue.pull_request && startsWith(github.event.comment.body, '/ci')` multiple times. Is that possible?

docs/README.md
rkiselenko: This works for `master` branch with `push` event. For another branch we need another badge.

.github/workflows/e2e.yml
rkiselenko: I'm not sure if it's good. Your suggestion downgrade readability.

mishoyama: @safronovD please add your thoughts on this

mishoyama (approved):

PR #676, opened 2021-12-23
[ISSUE-668] no delete AC before lvg removing  

### LINKED ISSUE #668: No need to remove AC on DeleteVolume request
**Describe the bug**
No need to remove AC on DeleteVolume request. 

**Environment :**
 - Version: 1.1.0

**To Reproduce**

1. Deploy statefulset with LVG. 
2. Delete statefulset.
3. Remove volume with lvg.

**Expected behavior**
After volume with lvg removing, corresponding AC was deleted, and recreated.

**Additional context**
Changes are needed [here](https://github.com/dell/csi-baremetal/blob/49cd1467eff3c6a1ce7dc272810ded66b8432f2e/pkg/common/volume_operations.go#L639-L645).

yurkov-anton:
### Resolves #668

Replace deletion AC with update AC before LVG removing.
Update contains setting size 0 and restore location.
Also add sync AC StorageClass in drive reconcile.

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [x] Choose PR label
- [ ] New unit tests added

## Testing
1. Deploy sts with hddlvg
2. Remove sts and pvc


rkiselenko (approved):

pkg/controller/capacitycontroller/capacitycontroller.go
```diff
@@ -151,6 +151,8 @@ func (d *Controller) createOrUpdateCapacity(ctx context.Context, drive api.Drive
 		// If ac is exists, update its size to drive size
 		if ac.Spec.Size != size {
 			ac.Spec.Size = size
+			// if size does not match, then probably AC was used for lvg
+			ac.Spec.StorageClass = util.ConvertDriveTypeToStorageClass(drive.GetType())
```
safronovD, line 155: Isn't better to check that `ac.Spec.StorageClass` not equals to `util.ConvertDriveTypeToStorageClass(drive.GetType())` and update if not?

pkg/common/volume_operations.go
```diff
@@ -635,10 +635,13 @@ func (vo *VolumeOperationsImpl) deleteLVGIfVolumesNotExistOrUpdate(lvg *lvgcrd.L
 	})
 
 	drivesUUIDs := vo.k8sClient.GetSystemDriveUUIDs()
-	// if only one volume remains - remove AC first and LogicalVolumeGroup then
+	// if only one volume remains - restore AC Location, and set 0 size for sync size with drive.
+	// Then delete LogicalVolumeGroup
 	if len(lvg.Spec.VolumeRefs) == 1 && !util.ContainsString(drivesUUIDs, lvg.Spec.Locations[0]) {
-		if err := vo.k8sClient.DeleteCR(context.Background(), ac); err != nil {
-			log.Errorf("Unable to delete AC %s: %v", ac.Name, err)
+		ac.Spec.Size = 0
```
mishoyama, line 641: Do we need to convert AC SC type here as well?

yurkov-anton: Then  need to request a disk, but size 0 and there is no need for this, And I think it's better to synch AC with the disk somewhere in one place.

pkg/controller/capacitycontroller/capacitycontroller.go
yurkov-anton: check added

pkg/common/volume_operations.go
mishoyama: We are going to delete LVG which means that AC will be converted back to non-LVG. May be it's more logical to do the conversion here and don't split the logic between two different places.

yurkov-anton: then logic for updating AC with disk will be in several places, isn't that worse? Here we not split, we only set location, otherwise,  we will not know with which disk update in the drive reconcile. Or we can set size and sc before together before lvg remove.

mishoyama: You set location to drive, but AC SC is HDDLVG. Support person will start looking at the non-existing LVG. This is confusing

misanthropicat (approved):

Aber4Nod (approved):

mitsaa (approved):

safronovD (approved):

mishoyama (approved):

rkiselenko:
something something

pkg/common/volume_operations_test.go
```diff
@@ -16,6 +16,39 @@ limitations under the License.
 
 package common
 
+import (
```
safronovD, line 19: import order

safronovD (approved):

pkg/common/volume_operations_test.go
yurkov-anton: done

rkiselenko (approved):

misanthropicat (approved):

PR #694, opened 2022-01-18
[ISSUE-690] disk affinity proposal 

### LINKED ISSUE #690: Proposal for disk affinity/anti-affinity feature
Current lvg storageclass cannot mount physical disks to PVs with the awareness of the requesting pods.

CSI should provide a way to support:
* mount PVs for the same pod on the same disk(disk affinity)
* do not mount PVs on the disks that already have some PVs from other specifc  pods.(disk anti-affinity) 

**Additional context**

Aber4Nod:
### Resolves #690 

Adding the proposal for disk affinity/anti-affinity feature

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [x] Choose PR label
- [ ] New unit tests added

## Testing
_Provide test details_

### DIFF
--- a/docs/proposals/disk-affinity.md
+++ b/docs/proposals/disk-affinity.md
@@ -214,3 +214,15 @@ over ACs in inputted VolumesPlanMap. \
   Disk affinity feature will work at Filter and Reserve stage. The difference with scheduler extender approach will consist in that, there will be only p3 described above (no need in p2). \
   Also, as there will be no AvailableCapacityReservation CRD, we will need to mark each PVC with annotation with reserved AC to it.
   So at controller side we will know which AC to use while creating the volume.
+* Storage capacity feature - in future it's planned to use storage capacity feature to simplify reservation logic and try to get rid of scheduler extender. \
+  In case of storage capacity usage - we can get rid of resources calculating and using this link of the filtering chain (look above for more information).
+  But we cannot get rid of scheduler extender/framework based scheduler completely for disk affinity feature usage, or we should write out reservation hooks by ourselves.
+  It is because of requirements nature: we should make a reserving decision by application factors (e.g. pod-label disk antiaffinity type).
+  So, basically there are following options:
+  * Use scheduler extender/framework based scheduler with storage capacity feature as we are doing currently (and discussed above in this proposal), but without resources calculations (calculations will be done while reporting the capacity as described in CSIStorageCapacity CRD).
+  * Use custom hooks for volume reservation and so do not use scheduler extender/framework based scheduler. This hook will be called by controller while CreateVolume API will be called.
+    This approach has its disadvantages because scheduler will have already scheduled the deployment at specific node - so a lot reschedule operations will be possible.
+    Also, there appear synchronous calls to this hook - which is an anti-pattern for controller/operator cases.
+  * Just make reservation at controller side while CreateVolume API will be called. This approach is similar to the previous but has no synchronous calls.
+    But instead it puts extra burden to controller, which should not be there. \
+  So, from them above options, remaining of scheduler extender/framework based scheduler looks more perspective and correct.
--- a/docs/proposals/disk-affinity.md
+++ b/docs/proposals/disk-affinity.md
@@ -17,21 +17,23 @@ CSI should provide a way to support:
 ## Proposal
 
 Usually for affinity/antiaffinity topology feature is used. But the minimal domain, on which this feature operates, is nodes, not disks.
-So we should use a different approach for this one. Another approach, which is proposed, is to use annotations for pointing out how to use disk affinity/antiaffinity:
+So we should use a different approach for this one. Another approach, which is proposed, is to use annotations for pointing out how to use disk affinity/antiaffinity: \
 affinity.volumes.csi-baremetal.dell.com/types: <pod-bound-required|pod-bound-preferred|volume-bound-required|volume-bound-preferred|dedicated-required|dedicated-preferred> - specify disk affinity type:
 * pod-bound-(required|preferred) - mount PVs for the same pod on the same disk, depending on the suffix - the behaviour is required/preferred;
 * volume-bound-(required|preferred) - place the specific PVs on the same disks, depending on the suffix - the behaviour is required/preferred;
 * dedicated-(required|preferred) - place volumes on the certain disks, which were dedicated for this application, depending on the suffix - the behaviour is required/preferred; 
-affinity.volumes.csi-baremetal.dell.com/volumes: "[pv-1, pv-2], [pv-3, pv4], ..." - specify which pvs should be placed on the same disks.
+
+affinity.volumes.csi-baremetal.dell.com/volumes: "[pv-1, pv-2], [pv-3, pv4], ..." - specify which pvs should be placed on the same disks. \
 affinity.volumes.csi-baremetal.dell.com/tolerations: nginx - specify acs for placement, which were dedicated to volumes of this pod.
 
 antiaffinity.volumes.csi-baremetal.dell.com/types: <pod-label-required|pod-label-preferred|volume-bound-required|volume-bound-preferred> - specify disk antiaffinity type:
 * pod-label-(required|preferred) - specify pods by their labels, which should not use the same drives as the current one, depending on the suffix - the behaviour is required/preferred;
 * volume-bound-(required|preferred) - specify which volumes should not be placed together on the same drives, depending on the suffix - the behaviour is required/preferred.
-antiaffinity.volumes.csi-baremetal.dell.com/labels: "${list of the labels}" - do not mount PVs on the disks that already have some PVs from other specific pods (specified by labels).
+
+antiaffinity.volumes.csi-baremetal.dell.com/labels: "${list of the labels}" - do not mount PVs on the disks that already have some PVs from other specific pods (specified by labels). \
 antiaffinity.volumes.csi-baremetal.dell.com/volumes: "[pv-1, pv-2], [pv-1, pv3], ..." - specify which pvs should not be placed on the same disks.
 
-If we want to dedicate drive/lvg for the certain application, then we should place label on it's ac:
+If we want to dedicate drive/lvg for the certain application, then we should place label on it's ac: \
 affinity.csi-baremetal.dell.com/taint: nginx
 
 ## Implementation
@@ -197,6 +199,6 @@ type VolumesPlanFilter interface {
 3.2. Implement planning filters for affinity and capacity (currently already implemented in planner.go logic). 
 3.3. Affinity planning filter can also be implemented as two separate filters (for affinity and antiaffinity) - as described at ChainOfResponsibility pattern.
 Depending on their type (affinity/antiaffinity) perform the corresponding filtering operations (e.g. for antiaffinity pod-label type - select requested pods by labels and filter out drives, used by them)
-over ACs in inputted VolumesPlanMap.
-3.4. Rename CapacityManager to Manager and iterating logic over planning filters to it. Before iteration - create the initial VolumesPlanMap structure with available AC at requested nodes.
+over ACs in inputted VolumesPlanMap. \
+3.4. Rename CapacityManager to Manager and iterating logic over planning filters to it. Before iteration - create the initial VolumesPlanMap structure with available AC at requested nodes. \
 3.5. Dynamically create needed filters (e.g. affinity filter only needed in case if affinity requests were set at DriveRequests) at reservation controller.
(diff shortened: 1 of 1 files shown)


rkiselenko (approved):

docs/proposals/disk-affinity.md
```diff
@@ -0,0 +1,161 @@
...
+* mount PVs for the same pod on the same disk (disk affinity);
+* do not mount PVs on the disks that already have some PVs from other specific pods (disk antiaffinity).
+
+## Proposal
+
+Usually for affinity/antiaffinity topology feature is used. But the minimal domain, on which this feature operates, is nodes, not disks.
+So we should use a different approach for this one. Another approach, which is proposed, is to use annotations for pointing out how to use disk affinity/antiaffinity:
+affinity.drives.csi-baremetal.dell.com/type: <pod-bound> - mount PVs for the same pod on the same disk;
+antiaffinity.drives.csi-baremetal.dell.com/type: <pod-label> - specify disk antiaffinity type: currently there is only one option - by pod labels;  
+antiaffinity.drives.csi-baremetal.dell.com/labels: "${list of the labels}" - do not mount PVs on the disks that already have some PVs from other specific pods (specified by labels).
```
safronovD, line 23: What about a case, if we don't want to allocate all PVs of one pod on the same LVG disk? For example, `www-1` and `www-1` from the sts below should be on the different disks?

Aber4Nod: Then, we should add another antiaffinity type, e.g. called by 'pod-bound', and just set it into antiaffinity type annotation:
antiaffinity.drives.csi-baremetal.dell.com/type: <pod-label|pod-bound>

docs/proposals/disk-affinity.md
```diff
@@ -0,0 +1,161 @@
...
+
+Current lvg storageclass cannot mount physical disks to PVs with the awareness of the requesting pods. So, it's possible
+for different applications to share the same disk.
+CSI should provide a way to support:
+* mount PVs for the same pod on the same disk (disk affinity);
+* do not mount PVs on the disks that already have some PVs from other specific pods (disk antiaffinity).
+
+## Proposal
+
+Usually for affinity/antiaffinity topology feature is used. But the minimal domain, on which this feature operates, is nodes, not disks.
```
mishoyama, line 19: Did you check other CSI drivers implementations? Especially vSAN CSI? Do they have similar logic?

docs/proposals/disk-affinity.md
```diff
@@ -0,0 +1,161 @@
...
+So we should use a different approach for this one. Another approach, which is proposed, is to use annotations for pointing out how to use disk affinity/antiaffinity:
+affinity.drives.csi-baremetal.dell.com/type: <pod-bound> - mount PVs for the same pod on the same disk;
```
mishoyama, line 21: Suggest to have key just `affinity.csi-baremetal.dell.com` without drives

mishoyama: - What if we want to have 2 PVs from a single pod to be placed on one disk and 1 on another?
- Are those rules are strict? Can we provide some API for user to not prevent scheduling when condition is not meet?
- Is `affinity.drives.csi-baremetal.dell.com/type` have only single key?

docs/proposals/disk-affinity.md
```diff
@@ -0,0 +1,161 @@
...
+antiaffinity.drives.csi-baremetal.dell.com/type: <pod-label> - specify disk antiaffinity type: currently there is only one option - by pod labels;
```
mishoyama, line 22: Ho will this work with other applications which don't have integration with proposed APIs?

docs/proposals/disk-affinity.md
```diff
@@ -0,0 +1,161 @@
...
+   ```
+   where NodeRequests - are requests by nodes, ReservationRequests - requests per volumes. \
+1.2. We should update it to support new type of requests: based on affinity annotations:
+```protobuf
+message AvailableCapacityReservation {
+   string Namespace = 1;
+   string Status = 2;
+   NodeRequests NodeRequests = 3;
+   repeated ReservationRequest ReservationRequests = 4;
+   DriveRequests DriveRequests = 5;
```
mishoyama, line 107: `AffinityRules` sounds better here

docs/proposals/disk-affinity.md
```diff
@@ -0,0 +1,161 @@
...
+           storageClassName: "csi-baremetal-sc-hddlvg"
+           resources:
+              requests:
+                 storage: 1Gi
+
+```
+
+#### Baremetal CSI Driver's improvements:
+
+At Baremetal CSI Driver we should improve Reservation flow (at reservation controller component) for supporting new type of volumes
```
mishoyama, line 87: How will this work on a filtering stage of the future standalone CSI scheduler component?

docs/proposals/disk-affinity.md
safronovD: Got it

docs/proposals/disk-affinity.md
```diff
@@ -0,0 +1,161 @@
...
+
+enum DriveAntiaffinityRequestType {
+   POD_LABEL = 1;
+}
+
+message DriveAntiaffinityPodLabelRequest {
+  repeated string Labels = 1;
+}
+```
+2. At scheduler extender (in case client set the drive affinity annotations) while creating capacity reservation - fill it with DriveRequests specified in p1.2 above.
```
safronovD, line 148: Would it work in case without deploying extender and ACR usage? For example, if we implement Storage Capacity feature. Link - https://kubernetes.io/docs/concepts/storage/storage-capacity/

docs/proposals/disk-affinity.md
Aber4Nod: Where can I read more about this: some proposals, planning?

mishoyama: https://github.com/dell/csi-baremetal/blob/master/docs/proposals/available-capacity-reservation.md

Aber4Nod: Yes, I've already known it. The question about standalone CSI scheduler, not about ACR.

docs/proposals/disk-affinity.md
Aber4Nod: It won't work it this case. But have we already proposals for supporting this feature? If we haven't so - it's better to write it first and then come up with this question.

docs/proposals/disk-affinity.md
Aber4Nod: Can you please give some examples of this kind of applications, which cannot support this API?

docs/proposals/disk-affinity.md
Aber4Nod: For the question of affinity naming - In future we may have some other affinity features, not related with drives. And it may become the problem if we exclude drive from naming here.

Aber4Nod: There wasn't such use case, but for this we can introduce new affinity type: 
affinity.drives.csi-baremetal.dell.com/type: volume-bound
Then we can have another annotation for affinity specifying which pvs should be placed on the same disks:
affinity.drives.csi-baremetal.dell.com/volumes: "[pv-1, pv-2], [pv-3, pv4]"

Likewise we can do the same for antiaffinity:
antiaffinity.drives.csi-baremetal.dell.com/type: volume-bound
And specify which drives should not be placed together on the same drives:
antiaffinity.drives.csi-baremetal.dell.com/volumes: "[pv-1, pv-2], [pv-1, pv3]"

Aber4Nod: For now affinity.drives.csi-baremetal.dell.com/type has single type. But in future it may have more types, as we've already discussed. So we may rename it to affinity.drives.csi-baremetal.dell.com/types

docs/proposals/disk-affinity.md
mishoyama: Some applications might be running by customer and we don't have control over it. The main concern here how will we guarantee that disk will in exclusive usage by application with proposed APIs.

docs/proposals/disk-affinity.md
mishoyama: > For the question of affinity naming - In future we may have some other affinity features, not related with drives. And it may become the problem if we exclude drive from naming here.

- LVG doesn't match to the single drive in common case since one LVG can be spread across multiple drives. Thus `drives` sounds confusing here. May use `volumes` instead if you think that just `affinity` is to common.

mishoyama: > There wasn't such use case, but for this we can introduce new affinity type: affinity.drives.csi-baremetal.dell.com/type: volume-bound Then we can have another annotation for affinity specifying which pvs should be placed on the same disks: affinity.drives.csi-baremetal.dell.com/volumes: "[pv-1, pv-2], [pv-3, pv4]"
> 
> Likewise we can do the same for antiaffinity: antiaffinity.drives.csi-baremetal.dell.com/type: volume-bound And specify which drives should not be placed together on the same drives: antiaffinity.drives.csi-baremetal.dell.com/volumes: "[pv-1, pv-2], [pv-1, pv3]"

please add this to proposal

mishoyama: > For required/preferred behaviour there are two possible solutions: We have modify our current annotations, e.g., in the following way: required.antiaffinity.drives.csi-baremetal.dell.com/labels: ecs-cluster-ss, ecs-cluster-pvg preferred.antiaffinity.drives.csi-baremetal.dell.com/labels: ecs-cluster-ss, ecs-cluster-pvg
> 
> Or we can modify type of affinity: antiaffinity.drives.csi-baremetal.dell.com/type: pod-label-required
> 
> The second solution may be more elegant

please add this to proposal

docs/proposals/disk-affinity.md
mishoyama: Currently in scheduler extender we use `filter` API to perform reservation which is not great, but it doesn't provide `reservation` API for us. But in case of scheduler this API will be exposed. So my question is about filtering stage in scheduler - what steps will be done there with the proposed API? Scheduler feature is not implemented yet, but want to avoid code rewriting for affinity/anti-affinity in the future.

docs/proposals/disk-affinity.md
mishoyama: Storage Capacity feature is also planned for implementation and it worth to consider this use case as well.

docs/proposals/disk-affinity.md
Aber4Nod: About vSAN CSI. All what I found up to now are the following annotations, which are placed directly at PVCs:
```
// best effort to spread PVs that have “foo-inst-6” as annotation on different disks. 
placement.beta.vmware.com/storagepool_antiAffinityPrefered : foo-inst-6 
```
The preferred policy asks for a best-effort anti-affinity, which means as far as possible all PVCs that have the annotation “placement.beta.vmware.com/storagepool_antiAffinityPrefered : foo-inst-6” should be placed on differnt vsan-direct disks, but PV creation should not fail if this anit-affinity can not be satisfied.
```
// make sure no to PVs with same PVC annotation go to the same disk
placement.beta.vmware.com/storagepool_antiAffinityRequired : foo-inst-6
```
If storagepool_antiAffinityRequired is used, then every PV with PVC annotation “placement.beta.vmware.com/storagepool_antiAffinityRequired : foo-inst-8” has to be placed on a different vSAN direct disk. Note that the node for the PV placement is still governed by where the corresponding POD gets placed, this anti-affinity rule only applies for PVs going to the same node.

Aber4Nod: topolvm hasn't this feature
direct-csi hasn't this feature
pmem-csi hasn't this feature
csi-driver-lvm hasn't this feature

docs/proposals/disk-affinity.md
Aber4Nod: If customer/we don't know which applications to select by the labels - we can make the following:
Mark drives/ac as dedicated, e.g. by adding the special label:
affinity.csi-baremetal.dell.com/dedicated: nginx
and then just use annotations at pods:
affinity.drives.csi-baremetal.dell.com/type: drive-label
affinity.drives.csi-baremetal.dell.com/drives: nginx

docs/proposals/disk-affinity.md
Aber4Nod: Added new types and options to proposal

docs/proposals/disk-affinity.md
Aber4Nod: Added new type and options to approach

docs/proposals/disk-affinity.md
Aber4Nod: Refactored

docs/proposals/disk-affinity.md
Aber4Nod: Added section about framework based scheduler to proposal

docs/proposals/disk-affinity.md
Aber4Nod: Added section about storage capacity feature to proposal

safronovD (approved):

PR #720, opened 2022-01-27
[ISSUE-717] Generic ephemeral volumes support

### LINKED ISSUE #717: Generic ephemeral volumes support
Need to support Ephemeral volumes in CSI. Changes must be done in Scheduler Extender to handle EphemeralVolumeSource case.

Code to be updated - https://github.com/dell/csi-baremetal/blob/16a14ae575d7273d13f851b13a212be5832165b7/pkg/scheduler/extender/extender.go#L222

safronovD:
### Resolves #717 

- Add ephemeral volume selecting in extender
- Add example

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [x] Choose PR label
- [x] New unit tests added

## Testing
_Provide test details_


pkg/scheduler/extender/extender.go
```diff
@@ -594,3 +609,7 @@ func (e *Extender) scNameStorageTypeMapping(ctx context.Context) (map[string]str
 	}
 	return scNameTypeMap, nil
 }
+
+func generateEphemeralVolumeName(podName, volumeName string) string {
```
rkiselenko, line 613: What the point of separate function ? Just use `fmt.Sprintf("%v-%v", podName, volume.Name)`

rkiselenko (approved):

pkg/scheduler/extender/extender.go
```diff
@@ -250,6 +250,21 @@ func (e *Extender) gatherCapacityRequestsByProvisioner(ctx context.Context, pod
 			}
 			continue
 		}
+		if v.Ephemeral != nil {
+			claimSpec := v.Ephemeral.VolumeClaimTemplate.Spec
+			if storageType, ok := scs[*claimSpec.StorageClassName]; ok {
+				storageReq, ok := claimSpec.Resources.Requests[coreV1.ResourceStorage]
+				if !ok {
+					ll.Errorf("There is no key for storage resource for PVC %s", v.Name)
```
mishoyama, line 258: Does it mean that incorrect SC name was passed?

mishoyama (commented):
Please update documentation - supported features list

pkg/scheduler/extender/extender.go
safronovD: We should have the same names during both ephemeral and generic ephemeral scenario. And for me, `requestName := generateEphemeralVolumeName(podName, volume.Name)` looks better, it marks that the whole name is not just a Volume name

pkg/scheduler/extender/extender.go
safronovD: No, it means that Volume size is not set, but from k8s specification it always must not be empty. Just check to not raise panic here, if we received incorrect data

safronovD:
> Please update documentation - supported features list

updated

mishoyama (approved):

pkg/scheduler/extender/extender.go
rkiselenko: But it implies from the code, this code tell you that the string is `podName - volume.Name` , there is no need of separate function for string formatting.

safronovD: In my opinion, it's not clear from the code, why we don't use VolumeName as it works with PVC. I've spent some time when just putting `v.Name` here and then decided to create a specific function with `EphemeralVolumes` in signature

safronovD: Go compiler should inline such simple functions, so it's just formatting question

PR #723, opened 2022-01-31
[ISSUE-713] SANITY_TEST parameter for sanity-ci

### LINKED ISSUE #713: Add test-sanity-ci target for k8s compatibility testing
Need to test CSI on as many as possible k8s versions, and at the same time need to spend as little as possible time for compatibility testing.

Proposed solution is to run as compatibility test only 1 test from e2e suite.

**Additional context**

misanthropicat:
### Resolves #713

SANITY_TEST value is parameterized from variables or from Jenkins job.

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [x] Choose PR label
- [ ] New unit tests added

## Testing
_Provide test details_


variables.mk
```diff
@@ -88,6 +88,7 @@ KIND_WAIT := 30s
 ### ci vars
 # timeout for short test suite, must be parsable as Go time.Duration (60m, 2h)
 SHORT_CI_TIMEOUT := 60m
+SANITY_TEST := [Testpattern: Dynamic PV (default fs)] volumes should store data
```
rkiselenko, line 91: Can we use `?=` instead of `:=` ?

rkiselenko (changes requested):

variables.mk
misanthropicat: Idk, actually I just declared this variable the same as others are declared. What's the advantage of your proposal?

rkiselenko: https://ftp.gnu.org/old-gnu/Manuals/make-3.79.1/html_chapter/make_6.html#SEC64

> If you'd like a variable to be set to a value only if it's not already set,

rkiselenko: `SANITY_TEST` will be used for running target locally `make SANITY_TEST='some focus' ...` while `SHORT_CI_TIMEOUT` mostly in CI pipelines.

rkiselenko (approved):

variables.mk
```diff
@@ -88,6 +88,7 @@ KIND_WAIT := 30s
 ### ci vars
 # timeout for short test suite, must be parsable as Go time.Duration (60m, 2h)
 SHORT_CI_TIMEOUT := 60m
+SANITY_TEST := volumes should store data  # focus expects to get a regexp as input not a simple string
```
mishoyama, line 91: Will that work without quotes?

misanthropicat: There is a problem with the following:
* if we pass variable from outside, it can't be with quotes as they processed inaccurately, so I'm adding quotes in Makefile during command construction
* if we have a variable in variables.mk with quotes, it'll be passed to Makefile with them, so we'll have doubled quotes
It doesn't work without quotes, I'm working on workaround currently.

safronovD (approved):

mishoyama (approved):

PR #738, opened 2022-02-07
[ISSUE-734] Proposal for drive exclusion feature

### LINKED ISSUE #734: Proposal for excluding specific drives feature
Some drives must be in use in block mode or reserved for other purposes. Need to provide API to blacklist them.

Come up with the proposal for excluding specific drives in CSI

safronovD:
### Resolves #734 

Add proposal

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [x] Choose PR label
- [ ] New unit tests added

## Testing
n/a


docs/proposals/excluding-drives.md
```diff
@@ -0,0 +1,105 @@
...
+## Abstract
+
+CSI Baremetal needs to have API for excluding specific drives from the common storage pool. 
+Excluded drives won't be used for allocating new Volumes or LVGs.
+
+## Background
+
+User stories:
+1. Deploy another CSI driver in kube-cluster.
+2. Reserve space for increasing root partition.
```
mishoyama, line 15: Not sure what does it mean...

docs/proposals/excluding-drives.md
```diff
@@ -0,0 +1,105 @@
...
+3. A kube-node is used for 2 or more kube-clusters (or not only for cluster).
```
mishoyama, line 16: Is that possible at all?

docs/proposals/excluding-drives.md
safronovD: To have one free drive in the system for a critical situation. Is it not applicable?

docs/proposals/excluding-drives.md
safronovD: Yes, but I think it is not a widely used scenario. Do you have other ideas why we need this feature?

mishoyama: I think that main use case that customer don't want to use some disks for whatever reason:
1. for another application not managed by CSI
2. as a workaround for any disk detection issues in CSI

safronovD: fixed

docs/proposals/excluding-drives.md
```diff
@@ -0,0 +1,105 @@
...
+  global:
+    drive_rule: ...
+  nodes:
+    - name: node1
+      enable_global: true/false
+      drive_rule:
+    - name: ...
+```
+
+#### Pass settings
```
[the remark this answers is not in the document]
safronovD: I proposed to use ConfigMap. Add info to clear it

mishoyama (approved):

misanthropicat (approved):

docs/proposals/excluding-drives.md
```diff
@@ -0,0 +1,108 @@
...
+- Serial Number
+- PID
+- VID
+```yaml
+drive_rule:
+  size:
+    less_then: 1Gi
+    more_then: 100Mi
+  # include drives with the following parameters only
+  in:
```
mitsaa, line 35: Do we really need to have "in" list? Maybe it would be better to have only "not_in" drives list? So, all the drives are acceptable by default, and drives from "not_in" list has to be excluded from CSI usage.

docs/proposals/excluding-drives.md
```diff
@@ -0,0 +1,108 @@
...
+Propose to process excluded drives on Discover stage in csi-baremetal-node part. 
+Node container will take drive setting from the ConfigMap created with csi-baremetal-deployment.
+
+Node ConfigMap can be updated in the following ways:
+1. Set global drive rule on CSI installation.
+
+Helm command for `csi-baremetal-driver` should be modified with:
+```yaml
+--set driver.node.global_drive_rule.size.more_then="1Gi" \
+--set driver.node.global_drive_rule.in.SN={sn1, sn2, sn3}
```
mitsaa, line 80: Did I get it right that a node will have only one "in" or/and "not_in" list, and every "--set" command will change that list(s)?
From my prospective, it would be beneficial to have add/remove commands to modify settings from that lists.

docs/proposals/excluding-drives.md
```diff
@@ -0,0 +1,108 @@
...
+```
+
+The option to modify rules for specific nodes is not supported due to complex formatting.
+
+2. After installing CSI user could edit a ConfigMap with CSI nodes setting to change included/excluded drives list
```
mitsaa, line 85: It's not clear to me how to take a look or display current settings of the ConfigMap. @safronovD, can you clarify that, if possible?

mitsaa (changes requested):

docs/proposals/excluding-drives.md
mishoyama: @mitsaa @rkiselenko it was my request to support include/exclude filtering options. We don't know all the use cases, but other CSI drivers work with **include** option only.

docs/proposals/excluding-drives.md
safronovD: > Did I get it right that a node will have only one "in" or/and "not_in" list, and every "--set" command will change that list(s)?
No, we can use both options together with different parameters. For example
```
in:
  type: HDD
not_in:
  SN: {sn1, sn2}
```
So CSI will use all HDD drives except ones with sn1 or sn2 serial numbers. But yes, correct config will be user responsibility.

docs/proposals/excluding-drives.md
```diff
@@ -0,0 +1,108 @@
...
+### User API
+
+The rules to include/exclude drives based on:
+- Size
+- Type (HDD, SSD, ...)
+- Serial Number
+- PID
+- VID
+```yaml
+drive_rule:
```
rkiselenko, line 30: Use `camelCase` not `snake_case` please.

rkiselenko (approved):
LGTM

docs/proposals/excluding-drives.md
safronovD: > From my prospective, it would be beneficial to have add/remove commands to modify settings from that lists.

These options are for installation via helm only. It doesn't work well with a list. So these rules will be empty and we just have an option to set a full list. I think it will be helpful only in the case when we know cluster architecture in automation

docs/proposals/excluding-drives.md
safronovD: Sure, updated

docs/proposals/excluding-drives.md
safronovD: Fixed

yurkov-anton (approved):

docs/proposals/excluding-drives.md
mitsaa: From my prospective, it is not clear what to do if a drive satisfies conditions from both filters "include" and "non-include".
If most of CSI drivers use one "include" option, it makes sense to follow common practice, I believe. Folks, can you please your views on this?

docs/proposals/excluding-drives.md
mitsaa: @safronovD , from previous experience, I'd strongly recommend to consider an ability to extend/modify lists/maps of drives which can be exposed/deprecated for CSI usage because we may have a huge panopticon of HW configuration once installation number has started growing. Can you please clarify, is that doable theoretically?

safronovD: In proposal 2 options are described:
- configure helm values on CSI installation
- editing setting ConfigMap to modify rules after installation

I guess your recomdadtion is cover by the second option

docs/proposals/excluding-drives.md
safronovD: Auto disk defining is the feature of our CSI, isn't it? :) In the case described below comment, we can have include/exclude rules for different option types (include all HDD, but exclude some drives via SN or PID). 
I agree that the rules struct here can bring some satisfies conditions, which will be user responsibility only. Could you advise on how we can clarify `drive_rule` to avoid it?

docs/proposals/excluding-drives.md
mitsaa: It looks we are done here.

docs/proposals/excluding-drives.md
mitsaa: If our CSI will be used on unrestricted HW configurations, we may have more issues than benefits with accepting to usage disks in auto mode. There are two ways here:
-  Accept all the disks excluding deprecated in "not_included" list/ConfigMap;
- Accept only disks from "included"  list/ConfigMap.
My tip is to have only **one** list/template of allowed **or** deprecated drives.

docs/proposals/excluding-drives.md
mitsaa: The only think I just want to point out. It's better to have user friendly interface than making a user to edit configs with possibility of creation incorrect configuration.

docs/proposals/excluding-drives.md
mishoyama: I think user will have to choose which strategy to follow:
1. Include rules
2. Exclude rules
Mixed mode should not be allowed

mitsaa: @mishoyama, two strategies looks cool. I am agree with that. 
BTW, is that possible to use something like regexp filters to set templates? I'm not sure that it would be useful to set definite VID/PID/SN-es for every node.

mitsaa (commented):
It looks we should make a decision about include/exclude strategies.

docs/proposals/excluding-drives.md
mishoyama: @mitsaa I suggest to start with the simple implementation until we get feedback from users with their use cases.

safronovD: Thank you! `driveRule` struct was updated. Please, check

mitsaa: OK, let's get feedback from users.

mitsaa (approved):

PR #746, opened 2022-02-10
[ISSUE-745] Wait for reservation to be confirmed

### LINKED ISSUE #745: Wait in scheduler extender for reservation to be confirmed

Currently we don't wait for reservation to be confirmed by Controller and return after ACR created.

Wait in scheduler extender for reservation to be confirmed until context is cancelled by timeout

rkiselenko:
### Resolves #745 

Wait for reservation to be confirmed,  continue on reading error,
return after reservation confirmed or wait context deadline

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [x] Choose PR label
- [x] New unit tests added

## Testing
_Provide test details_

### DIFF
--- a/pkg/scheduler/extender/extender.go
+++ b/pkg/scheduler/extender/extender.go
@@ -464,14 +464,14 @@ func (e *Extender) createReservation(ctx context.Context, ns, name string,
 				ll.Warningf("Failed to read capacity reservation '%s' CR", name)
 				continue
 			}
-			if acr.Spec.Status == v1.ReservationConfirmed {
+			if acr.Spec.Status == v1.ReservationConfirmed || acr.Spec.Status == v1.ReservationRejected {
 				ll.Infof("Capacity reservation '%s' status is '%s'", name, acr.Spec.Status)
 				readCh <- struct{}{}
 				return
 			}
 		}
 	}()
-	ll.Infof("Wait for capacity reservation '%s' become in status '%s'", name, v1.ReservationConfirmed)
+	ll.Infof("Wait for CR '%s' become in status '%s' or '%s'", name, v1.ReservationConfirmed, v1.ReservationRejected)
 	select {
 	case <-ctx.Done():
 		ll.Warningf("Context deadline reached capacity reservation '%s' CR", name)
--- a/pkg/scheduler/extender/extender.go
+++ b/pkg/scheduler/extender/extender.go
@@ -48,10 +48,6 @@ import (
 	annotations "github.com/dell/csi-baremetal/pkg/crcontrollers/node/common"
 )
 
-var (
-	reservationConfirmedDuration = 5 * time.Minute
-)
-
 // Extender holds http handlers for scheduler extender endpoints and implements logic for nodes filtering
 // based on pod volumes requirements and Available Capacities
 type Extender struct {
@@ -451,7 +447,6 @@ func (e *Extender) createReservation(ctx context.Context, ns, name string,
 	// or wait context deadline
 	readCh := make(chan struct{})
 	ticker := time.NewTicker(time.Second)
-	ctx, cancel := context.WithTimeout(ctx, reservationConfirmedDuration)
 	ll := e.logger.WithFields(logrus.Fields{
 		"method":      "createReservation",
 		"sessionUUID": ctx.Value(base.RequestUUID),
@@ -480,7 +475,6 @@ func (e *Extender) createReservation(ctx context.Context, ns, name string,
 		break
 	}
 	close(readCh)
-	cancel()
 	return nil
 }
 
--- a/pkg/scheduler/extender/extender.go
+++ b/pkg/scheduler/extender/extender.go
@@ -102,9 +102,7 @@ func (e *Extender) FilterHandler(w http.ResponseWriter, req *http.Request) {
 		}
 		return
 	}
-	defer func() {
-		_ = req.Body.Close()
-	}()
+	_ = req.Body.Close()
 
 	ll = ll.WithFields(logrus.Fields{
 		"pod": extenderArgs.Pod.Name,
@@ -169,9 +167,7 @@ func (e *Extender) PrioritizeHandler(w http.ResponseWriter, req *http.Request) {
 		ll.Errorf("Unable to decode request body: %v", err)
 		return
 	}
-	defer func() {
-		_ = req.Body.Close()
-	}()
+	_ = req.Body.Close()
 
 	ll.Info("Scoring")
 
@@ -217,9 +213,7 @@ func (e *Extender) BindHandler(w http.ResponseWriter, req *http.Request) {
 		}
 		return
 	}
-	defer func() {
-		_ = req.Body.Close()
-	}()
+	_ = req.Body.Close()
 
 	extenderBindingRes.Error = "don't know how to use bind API"
 	if err := resp.Encode(extenderBindingRes); err != nil {
--- a/pkg/scheduler/extender/extender.go
+++ b/pkg/scheduler/extender/extender.go
@@ -457,11 +457,11 @@ func (e *Extender) createReservation(ctx context.Context, ns, name string,
 		for range ticker.C {
 			acr := &acrcrd.AvailableCapacityReservation{}
 			if err := e.k8sClient.ReadCR(ctx, name, ns, acr); err != nil {
-				ll.Warningf("Failed to read capacity reservation '%s' CR", name)
+				ll.Warningf("Failed to read CR '%s' error: %s", name, err)
 				continue
 			}
 			if acr.Spec.Status == v1.ReservationConfirmed || acr.Spec.Status == v1.ReservationRejected {
-				ll.Infof("Capacity reservation '%s' status is '%s'", name, acr.Spec.Status)
+				ll.Infof("CR '%s' status is '%s'", name, acr.Spec.Status)
 				readCh <- struct{}{}
 				return
 			}
@@ -470,7 +470,7 @@ func (e *Extender) createReservation(ctx context.Context, ns, name string,
 	ll.Infof("Wait for CR '%s' become in status '%s' or '%s'", name, v1.ReservationConfirmed, v1.ReservationRejected)
 	select {
 	case <-ctx.Done():
-		ll.Warningf("Context deadline reached capacity reservation '%s' CR", name)
+		ll.Warningf("Context deadline reached CR '%s'", name)
 		break
 	case <-readCh:
 		break
--- a/pkg/scheduler/extender/extender.go
+++ b/pkg/scheduler/extender/extender.go
@@ -405,6 +405,7 @@ func getReservationName(pod *coreV1.Pod) string {
 	return namespace + "-" + pod.Name
 }
 
+// TODO https://github.com/dell/csi-baremetal/issues/747
 func (e *Extender) createReservation(ctx context.Context, ns, name string,
 	nodes []coreV1.Node, capacities []*genV1.CapacityRequest) error {
 	// ACR CRD
--- a/pkg/scheduler/extender/extender.go
+++ b/pkg/scheduler/extender/extender.go
@@ -441,14 +441,13 @@ func (e *Extender) createReservation(ctx context.Context, ns, name string,
 	// return after reservation confirmed
 	// or wait context deadline
 	readCh := make(chan struct{})
-	ticker := time.NewTicker(time.Second)
 	ll := e.logger.WithFields(logrus.Fields{
 		"method":      "createReservation",
 		"sessionUUID": ctx.Value(base.RequestUUID),
 	})
 	go func() {
-		defer ticker.Stop()
-		for range ticker.C {
+		for _, timer := range []time.Duration{1 * time.Second, 3 * time.Second, 10 * time.Second} {
+			time.Sleep(timer)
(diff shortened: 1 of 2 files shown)


misanthropicat (approved):

pkg/scheduler/extender/extender.go
```diff
@@ -431,6 +445,42 @@ func (e *Extender) createReservation(ctx context.Context, namespace string, name
 		// cannot create reservation
 		return err
 	}
+	// wait for reservation to be confirmed
+	// continue on reading error
+	// return after reservation confirmed
+	// or wait context deadline
+	readCh := make(chan struct{})
+	ticker := time.NewTicker(time.Second)
+	ctx, cancel := context.WithTimeout(ctx, reservationConfirmedDuration)
```
mishoyama, line 454: Don't context coming from scheduler without timeout? Why did you chose 5 minutes?

rkiselenko: > Don't context coming from scheduler without timeout? Why did you chose 5 minutes?

This context comes from [http handler](https://github.com/dell/csi-baremetal/blob/master/pkg/scheduler/extender/extender.go#L110) and there is no deadline or timeout on this context.

I chose 5 minutes because context must have deadline.

mishoyama: Do you mean the scheduler sends requests to the extender(s) and don't fail them by timeout?

pkg/scheduler/extender/extender.go
```diff
@@ -101,6 +106,9 @@ func (e *Extender) FilterHandler(w http.ResponseWriter, req *http.Request) {
 		}
 		return
 	}
+	defer func() {
+		_ = req.Body.Close()
```
mishoyama, line 110: Does it make sense to print error here?

pkg/scheduler/extender/extender.go
```diff
@@ -431,6 +445,42 @@ func (e *Extender) createReservation(ctx context.Context, namespace string, name
...
+	})
+	go func() {
+		defer ticker.Stop()
+		for range ticker.C {
+			acr := &acrcrd.AvailableCapacityReservation{}
+			if err := e.k8sClient.ReadCR(ctx, name, ns, acr); err != nil {
+				ll.Warningf("Failed to read capacity reservation '%s' CR", name)
+				continue
+			}
+			if acr.Spec.Status == v1.ReservationConfirmed {
```
mishoyama, line 467: I guess when it rejected it also make sense to return

pkg/scheduler/extender/extender.go
rkiselenko: I mean the context in this function doesn't have timeout at all.

pkg/scheduler/extender/extender.go
rkiselenko: If you only read from it, the error of Close is irrelevant.

pkg/scheduler/extender/extender.go
```diff
@@ -395,11 +409,11 @@ func getReservationName(pod *coreV1.Pod) string {
 	return namespace + "-" + pod.Name
 }
 
-func (e *Extender) createReservation(ctx context.Context, namespace string, name string, nodes []coreV1.Node,
-	capacities []*genV1.CapacityRequest) error {
+func (e *Extender) createReservation(ctx context.Context, ns, name string,
```
safronovD, line 412: We return nil after `createReservation` in `filter` function. Suitable nodes list is empty in this case. We should perform `handleReservation` after it became `RESERVED`
```
Feb 14 14:00:55.826 [INFO] [Extender] [createReservation] [0c2d8c1e-e46f-44bd-887c-226cde6a982d] Wait for capacity reservation 'volume-1563-baremetal-injector' become in status 'RESERVED'
Feb 14 14:00:56.829 [INFO] [Extender] [createReservation] [0c2d8c1e-e46f-44bd-887c-226cde6a982d] Capacity reservation 'volume-1563-baremetal-injector' status is 'RESERVED'
Feb 14 14:00:56.829 [INFO] [Extender] [FilterHandler] [baremetal-injector] [0c2d8c1e-e46f-44bd-887c-226cde6a982d] Construct response. Get 3 nodes in request. Among them suitable nodes count is 0. Filtered out nodes - map[]
```

safronovD (changes requested):

pkg/scheduler/extender/extender.go
rkiselenko: I dont get, please explain

pkg/scheduler/extender/extender.go
```diff
@@ -431,6 +445,42 @@ func (e *Extender) createReservation(ctx context.Context, namespace string, name
 		// cannot create reservation
 		return err
 	}
+	// wait for reservation to be confirmed
+	// continue on reading error
+	// return after reservation confirmed
+	// or wait context deadline
+	readCh := make(chan struct{})
+	ticker := time.NewTicker(time.Second)
```
safronovD, line 453: How many attempts are needed for one filter request? Should we make this timeout higher or increase in some time to not load k8s controller on large clusters?

pkg/scheduler/extender/extender.go
mishoyama: We can set any timeout from our side but it doesn't make a lot of sense if we aren't align with what scheduler does.

pkg/scheduler/extender/extender.go
rkiselenko: > How many attempts are needed for one filter request?

It depends

> Should we make this timeout higher or increase in some time to not load k8s controller on large clusters?

How much ?

pkg/scheduler/extender/extender.go
rkiselenko: Client can have own timeouts not aligned with server.

pkg/scheduler/extender/extender.go
safronovD: In `filter` function we read ACR and check it:
- if `IsNotFound` -> call `createReservation` and return nil, so we will response with empty list and `Filter Request` will be repeated by scheduler
- if found -> we call `handleReservation`, build and return suitable nodes list
In my opinion the goal is to process Filter Request only one time, so we need to call `handleReservation` after ACR is created and in RESERVED

pkg/scheduler/extender/extender.go
mishoyama: Not sure what do you mean. You have TCP session established which should be closed to avoid resources leak. If close operation failed error message must be logged.

pkg/scheduler/extender/extender.go
rkiselenko: The function returns `nil` and `handleReservation` called later then `CR` is in `RESERVED` state.

safronovD: Yes, after `FilterHandler` will be repeated. Why do we need it?

pkg/scheduler/extender/extender.go
safronovD: Not sure, need to have experiments. Could we have 10-20 sec timeout after some attempts?

pkg/scheduler/extender/extender.go
rkiselenko: Your note is related to `file` reading, but this is http body they're different.

The `req.Body.Close()` function returns an error in order to fulfill the `io.Closer` interface; the function doesn't necessarily return an error.

pkg/scheduler/extender/extender.go
rkiselenko: AFAIK `handleReservation` called after `CR` in the `RESERVED` state.

safronovD: But we don't call it after `createReservation`
```
	// read reservation
	reservation := &acrcrd.AvailableCapacityReservation{}
	err = e.k8sClient.ReadCR(ctx, reservationName, "", reservation)
	if err != nil {
		if k8serrors.IsNotFound(err) {
			// create new reservation
			if err := e.createReservation(ctx, pod.Namespace, reservationName, nodes, capacities); err != nil {
				// cannot create reservation
				return nil, nil, err
			}
			// not an error - reservation requested
			return nil, nil, nil
		}
		// issue with reading reservation
		return nil, nil, err
	}

	// reservation found
	return e.handleReservation(ctx, reservation, nodes)
```

pkg/scheduler/extender/extender.go
rkiselenko: Where this timeouts (10-20 sec) comes from ?

pkg/scheduler/extender/extender.go
rkiselenko: The issue is about wait context timeout, your notes is about logic refactoring it's out of this issue scope.

pkg/scheduler/extender/extender.go
safronovD: When I have increased RateLimiter Timeout, I made an experiment that `CreateVolume` request for a pod with 4 LVG volumes last 12.5 seconds, so the next ACR can became RESERVED after this period. It is applicable for baremetal Openshift cluster

rkiselenko: To make decision which timeout is most suitable we need better evidence(Tests and benchmarks with different cluster configurations and profiles.). I chose one second as start point.

pkg/scheduler/extender/extender.go
mishoyama: What is the goal of these changes then? It does make sense to wait for response only till scheduler is able to handle response from our side.

rkiselenko: Every http handler should have timeout so it wont stuck forever. I chose 5 minutes as a start point. Do you have any consideration how long it should be?

pkg/scheduler/extender/extender.go
mishoyama: Ok

pkg/scheduler/extender/extender.go
mishoyama: In this case what is the purpose to have 5 minutes timeout for `FilterHandler`?

rkiselenko: Because the nature of CSI, I assume some calls of `CR` creation can take much more than 5 seconds (with transition from requested to confirmed).

pkg/scheduler/extender/extender.go
safronovD: Could you please create Github issue and mention it in TODO in `filter` function?

pkg/scheduler/extender/extender.go
safronovD: I think, it might be configurable to avoid this situation

pkg/scheduler/extender/extender.go
rkiselenko: You better understand a problem here could you create an issue?

pkg/scheduler/extender/extender.go
rkiselenko: I don't know what we should denote in this issue. Any thoughts?‎

pkg/scheduler/extender/extender.go
safronovD: Sure

pkg/scheduler/extender/extender.go
mishoyama: When we patch scheduler configuration we override default `httpTimeout` from 5 to 15 seconds. It means that you don't need to add own timeout - just wait for http context to be cancelled.

pkg/scheduler/extender/extender.go
rkiselenko: I don't think its matter because context timeout is 15 seconds.

pkg/scheduler/extender/extender.go
rkiselenko: ok

pkg/scheduler/extender/extender.go
mishoyama: Agree that we need to have exponential back-off here to avoid Kube API overloading

rkiselenko: ok

safronovD: I thought the deadline timeout is 5 minutes... If it's just 15 seconds, it may be not critical

safronovD (approved):

yurkov-anton (approved):

safronovD:
Why do you replace `req.Close` from `defer` section?

rkiselenko:
> Why do you replace `req.Close` from `defer` section?

Because of coverage.

pkg/scheduler/extender/extender.go
```diff
@@ -431,6 +436,39 @@ func (e *Extender) createReservation(ctx context.Context, namespace string, name
...
+	// continue on reading error
+	// return after reservation confirmed
+	// or wait context deadline
+	readCh := make(chan struct{})
+	ll := e.logger.WithFields(logrus.Fields{
+		"method":      "createReservation",
+		"sessionUUID": ctx.Value(base.RequestUUID),
+	})
+	go func() {
+		for _, timer := range []time.Duration{1 * time.Second, 3 * time.Second, 10 * time.Second} {
```
mishoyama, line 449: If we change 15 seconds http timeout to something else will this code work correctly. May be it make sense to start with 1 second and multiple by 2 on each iteration and don't limit number of iterations and max value?

PR #755, opened 2022-02-21
[ISSUE-754]: adding retries on k8s operations

### LINKED ISSUE #754: Support retries on k8s operations
During scalability testing failed there are a lot of CIS Volumes in FAILED state due to the following error in the code:
```
Jan 19 08:40:32.334 [INFO] [CSINodeService] [NodeStageVolume] [pvc-01674695-3cc8-4dee-8392-140dd8250715] Volume status: CREATED

Jan 19 08:40:32.335 [DEBU] [CSINodeService] [NodeStageVolume] [pvc-01674695-3cc8-4dee-8392-140dd8250715] staging path is: /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-01674695-3cc8-4dee-8392-140dd82
50715/globalmount/dev

Jan 19 08:41:37.174 [ERRO] [CSINodeService] [NodeStageVolume] [pvc-01674695-3cc8-4dee-8392-140dd8250715] failed to get partition for volume {pvc-01674695-3cc8-4dee-8392-140dd8250715 103eee5d-9177-4609-bf07-c3c563780422 DRIVE HDD 51e15a0e-cd9b-469b-a768-1682b90ed95d [] 105906176 FS xfs GOOD OPERATIVE CREATED IN_USE false {} [] 0}: unable to find drive by location 103eee5d-9177-4609-bf07-c3c563780422
```

Since K8s API service is overloaded requests might time out. We don't handle them properly on NodeStageVolume request in GetVolumePath:
```
drive := d.crHelper.GetDriveCRByUUID(vol.Location) drive := d.crHelper.GetDriveCRByUUID(vol.Location)
 if drive == nil { return "", fmt.Errorf("unable to find drive by location %s", vol.Location) }
```

Also it was noticed that some other API calls are also timing out.

Add optional retries on k8s operations.

**Additional context**

Aber4Nod:
### Resolves #754 

Added optional retries on k8s operations.

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [x] Choose PR label
- [ ] New unit tests added

## Testing
_Provide test details_

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -24,7 +24,6 @@ import (
 	"net"
 	"net/http"
 	"strconv"
-	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
@@ -32,7 +31,6 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
-	grpcbackoff "google.golang.org/grpc/backoff"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
@@ -98,13 +96,7 @@ func main() {
 		logger.Fatalf("fail to create kubernetes client, error: %v", err)
 	}
 	kubeClient := k8s.NewKubeClient(k8SClient, logger, objects.NewObjectLogger(), *namespace,
-		backoff.NewExponentialHandler(&grpcbackoff.Config{
-			// TODO(n.mikhnenko): customize vars
-			BaseDelay:  30 * time.Millisecond,
-			Multiplier: 1.6,
-			Jitter:     0.5,
-			MaxDelay:   30 * time.Second,
-		}),
+		backoff.NewExponentialHandler(&backoff.DefaultConfig),
 	)
 	controllerService := controller.NewControllerService(kubeClient, logger, featureConf)
 	handler := util.NewSignalHandler(logger)
@@ -195,13 +187,7 @@ func createManager(ctx context.Context, client *k8s.KubeClient, log *logrus.Logg
 		}
 	}
 	wrappedK8SClient := k8s.NewKubeClient(client, log, objects.NewObjectLogger(), *namespace,
-		backoff.NewExponentialHandler(&grpcbackoff.Config{
-			// TODO(n.mikhnenko): customize vars
-			BaseDelay:  30 * time.Millisecond,
-			Multiplier: 1.6,
-			Jitter:     0.5,
-			MaxDelay:   30 * time.Second,
-		}),
+		backoff.NewExponentialHandler(&backoff.DefaultConfig),
 	)
 
 	kubeCache, err := k8s.InitKubeCache(ctx, log,
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -26,25 +26,25 @@ import (
 	"strconv"
 	"time"
 
-	grpcbackoff "google.golang.org/grpc/backoff"
 	"github.com/container-storage-interface/spec/lib/go/csi"
 	grpc_prometheus "github.com/grpc-ecosystem/go-grpc-prometheus"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/sirupsen/logrus"
 	"google.golang.org/grpc"
+	grpcbackoff "google.golang.org/grpc/backoff"
 	"k8s.io/apimachinery/pkg/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	ctrl "sigs.k8s.io/controller-runtime"
 
-	"github.com/dell/csi-baremetal/pkg/base/backoff"
 	acrcrd "github.com/dell/csi-baremetal/api/v1/acreservationcrd"
 	accrd "github.com/dell/csi-baremetal/api/v1/availablecapacitycrd"
 	"github.com/dell/csi-baremetal/api/v1/drivecrd"
 	"github.com/dell/csi-baremetal/api/v1/lvgcrd"
 	"github.com/dell/csi-baremetal/api/v1/volumecrd"
 	"github.com/dell/csi-baremetal/pkg/base"
+	"github.com/dell/csi-baremetal/pkg/base/backoff"
 	"github.com/dell/csi-baremetal/pkg/base/featureconfig"
 	"github.com/dell/csi-baremetal/pkg/base/k8s"
 	"github.com/dell/csi-baremetal/pkg/base/logger"
@@ -100,10 +100,10 @@ func main() {
 	kubeClient := k8s.NewKubeClient(k8SClient, logger, objects.NewObjectLogger(), *namespace,
 		backoff.NewExponentialHandler(&grpcbackoff.Config{
 			// TODO(n.mikhnenko): customize vars
-			BaseDelay:  30*time.Millisecond,
+			BaseDelay:  30 * time.Millisecond,
 			Multiplier: 1.6,
 			Jitter:     0.5,
-			MaxDelay:   30*time.Second,
+			MaxDelay:   30 * time.Second,
 		}),
 	)
 	controllerService := controller.NewControllerService(kubeClient, logger, featureConf)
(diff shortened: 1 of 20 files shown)


cmd/controller/main.go
```diff
@@ -94,7 +97,15 @@ func main() {
 	if err != nil {
 		logger.Fatalf("fail to create kubernetes client, error: %v", err)
 	}
-	kubeClient := k8s.NewKubeClient(k8SClient, logger, objects.NewObjectLogger(), *namespace)
+	kubeClient := k8s.NewKubeClient(k8SClient, logger, objects.NewObjectLogger(), *namespace,
+		backoff.NewExponentialHandler(&grpcbackoff.Config{
```
rkiselenko, line 101: What the reason to use `grpcbackoff` config here? In my opinion it's break `DI` and `grpcbackoff` must be in the `"github.com/dell/csi-baremetal/pkg/base/backoff"` package, config should be a part of `"github.com/dell/csi-baremetal/pkg/base/backoff"` package.

cmd/controller/main.go
```diff
@@ -94,7 +97,15 @@ func main() {
...
+			// TODO(n.mikhnenko): customize vars
```
rkiselenko, line 102: TODO must be denoted as gh issue with link in the comment.

pkg/base/backoff/handler.go
```diff
@@ -0,0 +1,47 @@
...
+type Handler interface {
+	Handle(retries int) time.Duration
+}
+
+type exponentialHandler struct {
+	config *backoff.Config
+}
+
+func (s *exponentialHandler) Handle(retries int) time.Duration {
+	// if there were no retries, then return baseDelay time
```
rkiselenko, line 22: It's good to point here what the default value of `baseDelay`

pkg/base/k8s/kubeclient.go
```diff
@@ -110,42 +115,86 @@ func (k *KubeClient) CreateCR(ctx context.Context, name string, obj k8sCl.Object
...
 	ll.Infof("Creating CR '%s': %s", crKind, k.objectsLogger.Log(obj))
-	err := k.Create(ctx, obj)
-	if err != nil {
-		if k8sError.IsAlreadyExists(err) {
-			ll.Infof("CR %s %s already exist", crKind, name)
-			return nil
+
+	opt := mergeKubeClientRequestOptions(opts...)
+	var retries int
+	for {
```
rkiselenko, line 121: Is this loop endless?

pkg/base/k8s/kubeclient.go
```diff
@@ -110,42 +115,86 @@ func (k *KubeClient) CreateCR(ctx context.Context, name string, obj k8sCl.Object
...
+		err := k.Get(ctx, k8sCl.ObjectKey{Name: name, Namespace: namespace}, obj)
+		if err == nil {
+			return nil
+		}
+
+		k.log.WithFields(logrus.Fields{"method": "ReadCR", "requestUUID": requestUUID.(string)}).
+			Errorf("error occurred while reading custom resource: '%v'", err)
+		if backoffErr := k.makeBackoffRetry(ctx, "ReadCR", requestUUID.(string), retries, *opt.MaxBackoffRetries); backoffErr != nil {
+			return errors.Wrap(backoffErr, fmt.Sprintf("backoff retries failed for origin error: '%v'", err.Error()))
```
rkiselenko, line 163: I think `%w` error verbs should be used here.

pkg/base/k8s/options.go
```diff
@@ -0,0 +1,28 @@
+package k8s
+
+var DefaultMaxBackoffRetries = 0
+
+type KubeClientRequestOptions struct {
```
rkiselenko, line 5: Can we have shorter naming ? `ClientOptions` instead of `KubeClientRequestOptions` for example or `DefaultOptions` for `DefaultKubeClientRequestOptions`. It's hard to read `Java` style naming in `go`.  Please use short, concise, evocative naming.

pkg/base/k8s/options.go
```diff
@@ -0,0 +1,28 @@
...
+	MaxBackoffRetries *int
+}
+
+func DefaultKubeClientRequestOptions() KubeClientRequestOptions{
+	return KubeClientRequestOptions{
+		MaxBackoffRetries: &DefaultMaxBackoffRetries,
+	}
+}
+
+func mergeKubeClientRequestOptions(opts ...*KubeClientRequestOptions) KubeClientRequestOptions {
```
rkiselenko, line 15: This line can be denoted for example as `func merge(opts ...*ClientOptions ) ClientOptions`, much readable IMO. The meaning of function can be implied from its arguments and return values.

pkg/common/volume_operations.go
```diff
@@ -276,7 +280,9 @@ func (vo *VolumeOperationsImpl) handleVolumeInProgress(ctx context.Context, log
 			log.Errorf("Timeout of %s for volume creation exceeded.", base.DefaultTimeoutForVolumeOperations)
 			volumeCR.Spec.CSIStatus = apiV1.Failed
 			// todo don't ignore error here
-			_ = vo.k8sClient.UpdateCRWithAttempts(ctx, volumeCR, 5)
+			_ = vo.k8sClient.UpdateCR(ctx, volumeCR, &k8s.KubeClientRequestOptions{
+				MaxBackoffRetries: &vo.updateCRAttempts,
```
rkiselenko, line 284: Why not log this error instead of ignoring?

rkiselenko (changes requested):

mishoyama:
@Aber4Nod do you plan to retry on every k8s API call when it failed by timeout? Does it make sense just to return timeout error and expect that k8s will retry later? It will dramatically simplify code on CSI side and minimize possible side effects.

cmd/controller/main.go
Aber4Nod: moved config

cmd/controller/main.go
Aber4Nod: used base config for this

pkg/base/k8s/kubeclient.go
Aber4Nod: fixed this

pkg/base/k8s/kubeclient.go
Aber4Nod: we have two errors here, %w is suitable only for 1 error cases

pkg/base/k8s/options.go
Aber4Nod: Renamed options

pkg/base/k8s/options.go
Aber4Nod: the functions are used to be named by the template: action + subject of action

pkg/common/volume_operations.go
Aber4Nod: logged this

Aber4Nod:
> @Aber4Nod do you plan to retry on every k8s API call when it failed by timeout? Does it make sense just to return timeout error and expect that k8s will retry later? It will dramatically simplify code on CSI side and minimize possible side effects.

Don't understand the comment. Can you please describe it in more details

pkg/base/backoff/handler.go
```diff
@@ -0,0 +1,48 @@
...
+	// randomize backoff delays so that if cluster of requests start at
+	// the same time, they won't operate in lockstep.
+	if curBackoff *= 1 + s.config.Jitter*(rand.Float64()*2-1); curBackoff < 0 {
+		return 0
+	}
+	return time.Duration(curBackoff)
+}
+
+// NewExponentialHandler is a constructor for backoff handler
+func NewExponentialHandler(config *Config) Handler {
```
rkiselenko, line 44: What do you think about a little changes I suggest here? 

```golang
// Handler is a backoff handler
type Handler interface {
	ExponentialHandle(retries int) time.Duration
}

type handler struct {
	config *Config
}
func (s *handler) ExponentialHandle(retries int) time.Duration {
	// some code here
}

func NewHandler(config *Config) Handler {
       return &handler{
		config: config,
	}
}
```

This code is much readable and expandable because only `ExponentialHandle` method doing exponential backoff, so we can introduce another method with for example `LinearHandle` with a little changes just in package.

rkiselenko (changes requested):

pkg/base/backoff/handler.go
Aber4Nod: We could do so, but in this case we should pass special configurations for each type of backoff (e.g. pass jitter to exponential only) and it would be excessive for caller methods to handle this information due to the nature of backoff method - it should be peripheral and easy to use.
Moreover by allowing to pass this configuration parameters for each run caller has ability to manage exponentiality of backoff - possible making it to not become exponential at all.
For now, as for me, it's better to remain logic as it is.

rkiselenko: You don't need special configuration, just extend default config. Dont see a problem with configuration could you show any example?

mitsaa (approved):

PR #765, opened 2022-03-03
[ISSUE-744]: CIS hardening support

### LINKED ISSUE #744: Support CIS hardening at Baremetal CSI
Currently CSI cannot be deployed with CIS hardening enabled. Need to support it at CSI. The version, which should be supported at rke - 1.6.

In order to support CIS it's need to integrate CSI with podSecurityPolicies and securityContexts (where it isn't implemented).

**Additional context**

Aber4Nod:
### Resolves #744 

Directly specifying UUIDs at dockerfiles for controller, node-controller components

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [x] Choose PR label
- [ ] New unit tests added

## Testing
_Provide test details_

### DIFF
--- a/pkg/controller/Dockerfile
+++ b/pkg/controller/Dockerfile
@@ -4,8 +4,8 @@ LABEL   description="Bare-metal CSI Controller Service"
 
 ADD     controller /controller
 
-RUN addgroup -S bmcsi && adduser -S bmcsi -G bmcsi
+RUN addgroup -S bmcsi && adduser -S bmcsi -u 1000 -G bmcsi
 
-USER bmcsi
+USER 1000
 
 ENTRYPOINT ["/controller"]
--- a/pkg/crcontrollers/node/Dockerfile
+++ b/pkg/crcontrollers/node/Dockerfile
@@ -4,8 +4,8 @@ LABEL   description="Baremetal CSI Operator"
 
 ADD     controller  node-controller
 
-RUN addgroup -S bmcsi && adduser -S bmcsi -G bmcsi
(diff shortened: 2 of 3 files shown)


safronovD (approved):

pkg/controller/Dockerfile
```diff
@@ -4,8 +4,8 @@ LABEL   description="Bare-metal CSI Controller Service"
 
 ADD     controller /controller
 
-RUN addgroup -S bmcsi && adduser -S bmcsi -G bmcsi
+RUN addgroup -S bmcsi && adduser -S bmcsi -u 1000 -G bmcsi
 
-USER bmcsi
+USER 1000
```
rkiselenko, line 9: This will set only user `UUID` and not `GID` why not use `bmcsi` user ?

Aber4Nod: If you ask, why this changes are needed - to pass cis hardening validation. Without directly specifying UUID it would not work.

rkiselenko: I just dont get why you use `UID` instead of name.

rkiselenko: adduser -u 1000 creates a user with id 1000 
if you're using just `USER` directive so `RUN` directive about is useless, correct me if I'm wrong.

Aber4Nod: There are some differences between this options. But more transparent way is to create and use the actual user.

Aber4Nod: No, in this case cis validation will fail

rkiselenko: ok, got it

rkiselenko (approved):

PR #768, opened 2022-03-10
[ISSUE-766] Remove replace directives

### LINKED ISSUE #766: Make csi-driver go.mod importable and remove replace directive as much as possible

Remove replace directive from sources

rkiselenko:
### Problem

Make csi-driver go.mod importable and remove replace directive as much as possible

In order to remove all replace we should remove deps which not related to actual driver. 

k8s e2e modules isn't importable and imported with replace directive.

### Resolution

Move all e2e tests from folder `test` to its own module `csi-baremetal-e2e-tests` and require `csi-baremetal` module as deps for it.

Cleanup deps and run `go mod tidy` remove all indirect deps.

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [x] Choose PR label
- [ ] New unit tests added

## Testing
_Provide test details_


tests/go.mod
```diff
@@ -0,0 +1,139 @@
...
+	github.com/onsi/gomega v1.14.0
+	github.com/sirupsen/logrus v1.8.1
+	gopkg.in/yaml.v2 v2.4.0
+	k8s.io/api v1.22.5
+	k8s.io/apimachinery v0.23.4
+	k8s.io/client-go v0.23.4
+	k8s.io/kubernetes v1.22.5
+)
+
+require (
```
safronovD, line 21: Why we have 2 `require` sections here?

go.mod
```diff
@@ -5,63 +5,25 @@ go 1.16
 require (
```
safronovD, line 5: Should we change go version here too?

tests/e2e/reports/report.xml
```diff
@@ -0,0 +1,543 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
yurkov-anton, line 1: Is this file need to merge PR?

yurkov-anton (approved):

tests/go.mod
rkiselenko: idk go mod works in that way

go.mod
rkiselenko: No

tests/go.mod
safronovD: thank you for the reference

go.mod
safronovD: So we have go 1.16 in devkit and in csi-beremetal go mod, go 1.17 in PR validation and in csi-baremetal-tests go.mod. Why do we need to upgrade go version for tests and why can't update it everywhere?

rkiselenko: `go help go.mod` 

> The expected language version, set by the go directive, determines
> which language features are available when compiling the module.
> Language features available in that version will be available for use.
> Language features removed in earlier versions, or added in later versions,
> will not be available. Note that the language version does not affect
> build tags, which are determined by the Go release being used.

rkiselenko: > The basic guideline is fairly simple: a "go" directive 1.N means that the code in that module is permitted to use language features that existed in 1.N even if those features were removed in later releases of the language.

safronovD: Ok, could you please explain why do we need to upgrade go version for csi-baremetal-tests only? Is it works in devkit with go 1.16?

rkiselenko: There is no "version upgrade", `go 1.N` declare the version as described above, you free to use any version.

safronovD: But we use go compailer 1.16 installed in devkit and declare go 1.17 in go.mod. May we have conflicts?

rkiselenko: no

safronovD (approved):

Makefile
```diff
@@ -92,33 +92,36 @@ install-protoc:
 install-compile-proto: install-protoc compile-proto
 
 install-controller-gen:
-	# Build controller-gen from go.mod
```
safronovD, line 95: Could you please remove `tidy` target too?

rkiselenko: I dont get.

PR #875, opened 2022-06-06
[ISSUE-856] report non default namespace drive event

### LINKED ISSUE #856: When csi-baremetal is deployed on non-default namespace, the CSI baremetal application generates events in the default namespace
**Describe the bug**
When csi-baremetal is deployed in non-default namespace and when the disk fails, the csi-baremetal application generates events in the default namespace.

 - Hardware: Any
 - OS: Any
 - Version: 1.1.0

**To Reproduce**
1. Deploy csi-baremetal in deployed in non-default namespace
2. Disk failure
3. kubectl get events -A to check the DriveHealthFailure  event on default namespace.

**Expected behavior**
When csi-baremetal is deployed in non-default namespace and when the disk fails, the csi-baremetal application generates events in the non-default namespace.

**Screenshots**
[root@centos76 ~]# kubectl get events -A | grep 40S0A0KRF1QF
default 6s Error DriveHealthFailure drive/09275394-45ed-426f-a735-0aaaf67ab326 Drive health is: BAD, previous state: GOOD. Drive Details: SN='40S0A0KRF1QF', Model='ATA TOSHIBA MG06ACA8', Type='HDD', Size='8001000000000', Firmware='GA08', Slot='21', NodeName='master0.ocp4.carnelian.com'
[root@centos76 ~]#
**Additional context**

yanxiucai:
Signed-off-by: yanxiucai <Yanxiu_Cai@dell.com>

### Resolves #856 

Update the code in https://github.com/dell/csi-baremetal/blob/master/pkg/events/recorder/recorder.go function makeEvent() to generate events for drives in the same namespace where CSI is installed.
## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [x] Choose PR label
- [x] New unit tests added

## Testing
[test.txt](https://github.com/dell/csi-baremetal/files/8844036/test.txt)

### DIFF
--- a/pkg/events/recorder/recorder.go
+++ b/pkg/events/recorder/recorder.go
@@ -20,6 +20,7 @@ package recorder
 import (
 	"fmt"
 	"math/rand"
+	"os"
 	"sync"
 	"time"
 
@@ -67,7 +68,12 @@ func (sr *SimpleRecorder) makeEvent(ref *v1.ObjectReference, labels map[string]s
 	t := metav1.NewTime(now)
 	namespace := ref.Namespace
 	if namespace == "" {
-		namespace = metav1.NamespaceDefault
+		if key := os.Getenv("NAMESPACE"); key != "" {
+			namespace = os.Getenv("NAMESPACE")
+			ref.Namespace = namespace
+		} else {
+			namespace = metav1.NamespaceDefault
+		}
 	}
 	return &v1.Event{
 		ObjectMeta: metav1.ObjectMeta{
--- a/pkg/events/recorder/recorder.go
+++ b/pkg/events/recorder/recorder.go
@@ -67,11 +67,15 @@ func (sr *SimpleRecorder) makeEvent(ref *v1.ObjectReference, labels map[string]s
 	}
 	t := metav1.NewTime(now)
 	namespace := ref.Namespace
-	if namespace == "" {
+	if ref.Kind == "Drive" {
 		if key := os.Getenv("NAMESPACE"); key != "" {
-			namespace = os.Getenv("NAMESPACE")
+			namespace = key
 			ref.Namespace = namespace
 		} else {
+			namespace = metav1.NamespaceDefault
+		}
+	} else {
+		if namespace == "" {
 			namespace = metav1.NamespaceDefault
 		}
 	}
--- a/pkg/events/recorder/recorder_test.go
+++ b/pkg/events/recorder/recorder_test.go
@@ -167,6 +167,43 @@ func TestSimpleRecorder_Eventf(t *testing.T) {
 				Type:           "Health",
 			},
 		},
+		{
+			name: "drive event when namespace ENV is not setting in csi pod",
+			fields: fields{
+				sink:   new(mocks.EventSink),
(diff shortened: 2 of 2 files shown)


mishoyama:
@yanxiucai please fix `codecov/patch` check

pkg/events/recorder/recorder.go
```diff
@@ -67,7 +68,12 @@ func (sr *SimpleRecorder) makeEvent(ref *v1.ObjectReference, labels map[string]s
 	t := metav1.NewTime(now)
 	namespace := ref.Namespace
 	if namespace == "" {
-		namespace = metav1.NamespaceDefault
+		if key := os.Getenv("NAMESPACE"); key != "" {
```
mishoyama, line 71: CSI generates events for cluster scoped CRs like `drives` and namespaced CRs like `volumes`.  How will that work when CSI in installed in non-default namespace, but volume CR is in default?

yanxiucai: I think we need to add a check here, it is for drives only.

pkg/events/recorder/recorder.go
```diff
@@ -67,7 +68,12 @@ func (sr *SimpleRecorder) makeEvent(ref *v1.ObjectReference, labels map[string]s
...
+			namespace = os.Getenv("NAMESPACE")
```
libzhang, line 72: no need to use os.Getenv to retrieve NAMESPACE again,  use 
```
namespace = key  
```
is enough.

yanxiucai: Yes.

pkg/events/recorder/recorder.go
yanxiucai: updated

mishoyama: Sorry, my comment was wrong. For cluster scoped CRs namespace will be empty, for namespaced - non-empty. @safronovD am I correct?

yanxiucai: ref.Nmespace here is empty or "default" for the namespaced CR while it is in default namespace?

safronovD: In my opinion, the first implementation was correct - `namespace == ""` means that resource is cluster-scoped, so we generate event for it in CSI NS and it doesn't matter is it `Drive` or something else

vgupta-mickey (commented):
I am not 100% how to resolve this issue. However,  when you generate event on a cluster-scoped resource, the InvolvedObject.Name = cluster-scoped eresource and InvolvedObject.Namespace = "". 
The Event.ObjectMeta.Namespace = default.

yanxiucai:
> However, when you generate event on a cluster-scoped resource, the InvolvedObject.Name = cluster-scoped eresource and InvolvedObject.Namespace = "".
> The Event.ObjectMeta.Namespace = default.

@vgupta-mickey In original implement, it always generate the event on cluster-scoped resource (drive) in default namespace but not in the namespace that CSI is installed.

yanxiucai:
Close this PR as no code change is needed in CSI side.

PR #887, opened 2022-06-14
[ISSUE-884] device or resource busy when NodeStageVolume

### LINKED ISSUE #884: CSI failed to deploy HDD volumes after upgrade when stage volume
**Describe the bug**
below logs is reported in csi node log,
```
MountVolume.MountDevice failed for volume "pvc-b8dd3c41-7ca1-4377-a1ee-2be8d9876c28" : rpc error: code = Internal desc = failed to stage volume: unable to find part name for device /dev/sdh by uuid b8dd3c41-7ca1-4377-a1ee-2be8d9876c28: exit status 1
```
detail logs,
```
Jun  1 15:05:47.327 [INFO] [CSINodeService] [NodeStageVolume] [pvc-b8dd3c41-7ca1-4377-a1ee-2be8d9876c28] Processing request
 1185 Jun  1 15:05:47.345 [INFO] [CSINodeService] [NodeStageVolume] [pvc-b8dd3c41-7ca1-4377-a1ee-2be8d9876c28] Volume status: PUBLISHED
 1186 Jun  1 15:05:47.345 [DEBU] [CSINodeService] [NodeStageVolume] [pvc-b8dd3c41-7ca1-4377-a1ee-2be8d9876c28] staging path is: /var/lib/kubelet/plugins/kuber      netes.io/csi/pv/pvc-b8dd3c41-7ca1-4377-a1ee-2be8d9876c28/globalmount/dev
 1187 Jun  1 15:05:47.347 [DEBU] [DriveProvisioner] [GetVolumePath] [pvc-b8dd3c41-7ca1-4377-a1ee-2be8d9876c28] Got drive &{TypeMeta:{Kind: APIVersion:} Object      Meta:{Name:03f75146-d83f-4a54-9c96-61a01ef5e7a5 GenerateName: Namespace: SelfLink:/apis/csi-baremetal.dell.com/v1/drives/03f75146-d83f-4a54-9c96-61a01ef      5e7a5 UID:d97ab3a5-cd30-4ea9-847a-0d31d54f3b3b ResourceVersion:6043135 Generation:910 CreationTimestamp:2022-05-27 07:23:43 +0000 UTC DeletionTimestamp:      <nil> DeletionGracePeriodSeconds:<nil> Labels:map[app:csi-baremetal app.kubernetes.io/name:csi-baremetal] Annotations:map[] OwnerReferences:[] Finalizer      s:[] ClusterName: ManagedFields:[{Manager:node Operation:Update APIVersion:csi-baremetal.dell.com/v1 Time:2022-06-01 10:38:06 +0000 UTC FieldsType:Field      sV1 FieldsV1:{"f:metadata":{"f:labels":{".":{},"f:app":{},"f:app.kubernetes.io/name":{}}},"f:spec":{".":{},"f:Firmware":{},"f:Health":{},"f:NodeId":{},"      f:PID":{},"f:Path":{},"f:SerialNumber":{},"f:Size":{},"f:Slot":{},"f:Status":{},"f:Type":{},"f:UUID":{},"f:Usage":{},"f:VID":{}}} Subresource:}]} Spec:{      UUID:03f75146-d83f-4a54-9c96-61a01ef5e7a5 VID:ATA PID:TOSHIBA MG06ACA8 SerialNumber:90D0A0E0F1QF Health:GOOD Type:HDD Size:8001000000000 Status:ONLINE U      sage:IN_USE NodeId:75d97d13-3480-4f5f-b49b-f8e2d95ac7fe Path:/dev/sdh Enclosure: Slot:18 Bay: Firmware:GA08 Endurance:0 LEDState: IsSystem:false IsClean      :false XXX_NoUnkeyedLiteral:{} XXX_unrecognized:[] XXX_sizecache:0}}
 1188 Jun  1 15:05:47.347 [DEBU] [DriveProvisioner] [GetVolumePath] [pvc-b8dd3c41-7ca1-4377-a1ee-2be8d9876c28] Got device /dev/sdh
 1189 Jun  1 15:05:47.347 [DEBU] [PartitionOperations] [SearchPartName] [b8dd3c41-7ca1-4377-a1ee-2be8d9876c28] Search partition number for device /dev/sdh and       uuid b8dd3c41-7ca1-4377-a1ee-2be8d9876c28
 1190 Jun  1 15:05:47.348 [ERRO] [Executor] [blockdev --rereadpt -v /dev/sdh] [795.294µs] [795294] [pkg/base/command/cmdexec.go:186] [runCmdFromCmdObj] stdout      : , stderr: blockdev: ioctl error on BLKRRPART: Device or resource busy, Error: exit status 1
```

 - Hardware: [Dell PowerEdge]
 - OS: [sles15]
 - Version: [1.1.0]

**To Reproduce**
deploy HDD volumes after upgrade

**Expected behavior**
no error reported and node volume stage success.;

**Screenshots**

**Additional context**

libzhang:
### Resolves #884 
blockdev --rereadpt -v <device> return "device or resource busy" when related volume was already mounted.(volume in published status). 
partprobe did not have this issue,  but due to known issues [udev issue](https://github.com/dell/csi-baremetal/blob/master/docs/kernel-distribution-dependenant-tools.md), 
we propose another solution:
in NodeStageVolume function,  when GetVolumePath,  for published volume, we just skip partition table sync.  and all other flows(get part uuid etc.) will not affect. 

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [x] Choose PR label
- [x] New unit tests added

## Testing
_Provide test details_


pkg/node/provisioners/drive_provisioner.go
```diff
@@ -224,7 +224,15 @@ func (d *DriveProvisioner) GetVolumePath(vol *api.Volume) (string, error) {
 	}
 	volumeUUID, _ = util.GetVolumeUUID(volumeUUID)
 
-	partNum, err := d.partOps.SearchPartName(device, volumeUUID)
+	// For volumes that are in Published status, ignore partition re-read and
+	// get the partition number directly.
+	var partNum string
+	if vol.CSIStatus == apiV1.Published {
+		partNum, err = d.partOps.GetPartitionNameByUUID(device, volumeUUID)
```
rkiselenko, line 231: Please don't ignore errors here, it is not clear which call the error handling below belongs to

rkiselenko (changes requested):

pkg/node/provisioners/drive_provisioner.go
libzhang: I assume if/else block can combine to one error handling,  the purpose is to get partition name.  so the error handleing below belongs to the above if/else block. 
do you mean we should logging  2 different error logs in if/else ?

rkiselenko: one call one error handling

libzhang: how about below logic,  is it ok ? 
```
229     var handleFunc func(string, string) (string, error)  
230     if vol.CSIStatus == apiV1.Published {  
231         handleFunc = d.partOps.GetPartitionNameByUUID  
232     } else {  
233         handleFunc = d.partOps.SearchPartName                        
234     }     
235           
236     partNum, err := handleFunc(device, volumeUUID)   
237                                                      
238     if err != nil {  
239         // on device disconnect or node reboot device name might change and we need to re-sync drive info   
240         return "", fmt.Errorf("unable to find part name for device %s by uuid %s: %w", device, volumeUUID, err)  
241     }             
```

pkg/node/provisioners/drive_provisioner.go
```diff
@@ -224,7 +224,15 @@ func (d *DriveProvisioner) GetVolumePath(vol *api.Volume) (string, error) {
...
+	} else {
+		partNum, err = d.partOps.SearchPartName(device, volumeUUID)
```
safronovD, line 233: We have a case when a volume in PUBLISHED state but not really mounted - after physical node reboot kubelet doesn't send UnStage and UnPublish requests and when node is back we receive NodeStage request for PUBLISHED volume (and have to bind/mount it again)
Could you please clarify a difference between `partprobe` and `blockdev` usage - why we have an error here? And could we use another tool for partition syncing but without udev calls?

libzhang: hi @safronovD,   I think the fix here will handle the case you described properly.  it  just does not re-read (resync) the kernel partition table again.  it still will get the part uuid directly and then do subsequent mount operations. 

for the difference b/w partprobe and blockdev,  I see blockdev will use the ioctl (BLKPART)  syscall call to do related operation,  while partprobe not use the same one and I found a link that said partprobe indeed did some improvment for mounted device. (https://bugs.freedesktop.org/show_bug.cgi?id=20813#:~:text=This%20is%20caused%20by%20the%20BLKRRPART%20ioctl%20failing,is%20really%20just%20a%20bug%2Ffeature%20in%20the%20kernel)

regarding the tool,  I do some research and found cfdisk/sfdisk,  but they all use the same api as blockdev(ioctl BLKPART)  and will get the same error.

pkg/node/provisioners/drive_provisioner.go
rkiselenko: Checkout example below: 
```golang
var partNum string
if partNum, err = d.partOps.SearchPartName(device, volumeUUID); err != nil {
  return "", fmt.Errorf("search part name for %s by uuid %s: %w", device, volumeUUID, err)
}
if vol.CSIStatus == apiV1.Published {
  if partNum, err = d.partOps.GetPartitionNameByUUID(device, volumeUUID); err != nil {
    return "", fmt.Errorf("get part name by uuid %s for device:%s : %w", volumeUUID, device, err)
  } 
}
```

libzhang: thank you @rkiselenko . changed the code, please help review again.

libzhang:
@safronovD @rkiselenko,  may you help review the code ?

safronovD:
@yurkov-anton Could you please take a look?

pkg/node/provisioners/drive_provisioner.go
libzhang: thanks @yurkov-anton for the review. 
do you mean skip blockdev error "device or resource busy" in underlying SearchPartName function ?
then it will need many code changes to pass the volume CSI status to the function, now the search PartName did not care about the volume,  just the uuid. 
another thing is "device or resource busy" is just the blockdev command stdout,  the err returned is "exit status -1",  so it is not easy to judge this.

yurkov-anton: Can we check for "Block device is busy"  in a error returned by `SearchPartName`?
Skiping synch of part table can be cause undefined behavior when disk was changed or disconnected.

libzhang: Hi @yurkov-anton,  in current code,  if SearchPartName returns error,  then no partuuid is returned, and subsequent action such as mounting the volume could not be done.  that's the reason I do a replacement for only get partuuid if volume is already in published status.  you can check the code call chain as below,  if we did not skip the partition re-read in such scenario, then it will fall into such  
error loop as now the problem describeed, 
```
   NodeStageVolume
        GetVolumePath
            SearchPartName
                SyncPartitionTable
```
here why we skip partition resync(partition re-read to notify the kernel) , since volume was already in published status, that indicated the partition was already be notified before,  even the node was rebooted,  that partition would not change.  we just need to re-read the partition uuid, no need to notify the kernel,  right?

libzhang: Hi @yurkov-anton,  sorry to bother you again,  could you help review again? any suggestions for my above comments ?

pkg/node/provisioners/drive_provisioner.go
libzhang: Hi @rkiselenko,  changed the code as your suggestions,  could you please help review ?

pkg/node/provisioners/drive_provisioner.go
```diff
@@ -224,10 +224,19 @@ func (d *DriveProvisioner) GetVolumePath(vol *api.Volume) (string, error) {
...
-		// on device disconnect or node reboot device name might change and we need to re-sync drive info
-		return "", fmt.Errorf("unable to find part name for device %s by uuid %s: %w", device, volumeUUID, err)
+	// For volumes that are in Published status, ignore partition re-read and
+	// get the partition number directly.
+	var partNum string
+	if vol.CSIStatus == apiV1.Published {
+		if partNum, err = d.partOps.GetPartitionNameByUUID(device, volumeUUID); err != nil {
+			return "", fmt.Errorf("get part name by uuid %s for device:%s : %w", volumeUUID, device, err)
+		}
+	} else {
```
rkiselenko, line 234: The else part is excess here

rkiselenko (changes requested):

pkg/node/provisioners/drive_provisioner.go
libzhang: Hi @rkiselenko ,   the logic here is for published volume,  we just get the partition uuid directly,  and for others,  we will use SearchPartName to re-sync partition table and then get the partition uuid.  so here the else part is necessary.   we need to pick one of them to execute based on the volume status. 
is my logic correct ?

pkg/node/provisioners/drive_provisioner.go
yurkov-anton: Hi @libzhang, sorry for the long answer.
I've told about ignore synch error only if message contain "Disk busy"  and volume is PUBLISHED, because this is expected error in this way. 
```
partName, err = SearchPartName
if err {
    if "Disk is busy" in err and vol.CSIStatus == apiV1.Published {
        partName, err = GetPartitionNameByUUID
   }
   if err {
         return err
   }
}
```
We have case that @safronovD was talking about:
1. physical node reboot
2. something wrong with disk or removed physically drive
3. VolumeCR still published and OS keep path to disk.
4. node is back and NodeStage send
5. CSI get part path by SearchPartName, but path is missing

I think can be another cases, which we do not take into account. We need to synch part table, if we can, before use part path.

pkg/node/provisioners/drive_provisioner.go
yurkov-anton: @rkiselenko сould you please clarify why this is excess part?

pkg/node/provisioners/drive_provisioner.go
libzhang: thanks @yurkov-anton.  I got your point.  will work on this way.

libzhang: Hi @yurkov-anton,  changed the code based on your comments,  may you help review ? thanks.

pkg/node/provisioners/drive_provisioner.go
```diff
@@ -227,14 +228,16 @@ func (d *DriveProvisioner) GetVolumePath(vol *api.Volume) (string, error) {
...
 	var partNum string
-	if vol.CSIStatus == apiV1.Published {
-		if partNum, err = d.partOps.GetPartitionNameByUUID(device, volumeUUID); err != nil {
-			return "", fmt.Errorf("get part name by uuid %s for device:%s : %w", volumeUUID, device, err)
-		}
-	} else {
-		if partNum, err = d.partOps.SearchPartName(device, volumeUUID); err != nil {
+	if partNum, err = d.partOps.SearchPartName(device, volumeUUID); err != nil {
+		// blockdev command run return exec.ExitError
+		if _, ok := err.(*exec.ExitError); ok && vol.CSIStatus == apiV1.Published {
```
yurkov-anton, line 233: blockdev can return a few type of errors, we need to skip only disk is busy

libzhang: Hi  @yurkov-anton,  now I checked the exec_posix.go code,   we can use the exit status code to judge this instead of stderr output.   the exitError.ProcessState.ExitStatus will catch the exit status code. 
do you think it is ok ?

yurkov-anton: As I know blockdev exit code doesn't say about type of error.
`err` contains error massage from blockdev. And we can check that `err` contains `Device or resource busy` part.

libzhang: Thank you @yurkov-anton for the reminder,  I found exitError contains Stderr, that's what we need here. updated code,  please help take a look again.

pkg/node/provisioners/drive_provisioner.go
```diff
@@ -224,10 +226,23 @@ func (d *DriveProvisioner) GetVolumePath(vol *api.Volume) (string, error) {
...
-		return "", fmt.Errorf("unable to find part name for device %s by uuid %s: %w", device, volumeUUID, err)
+	// For volumes that are in Published status, ignore partition re-read and
+	// get the partition number directly.
+	var partNum string
+	if partNum, err = d.partOps.SearchPartName(device, volumeUUID); err != nil {
+		// blockdev command run return exec.ExitError, exit error code 1 means
+		// the blockdev hit "device or resource busy" error.
+		if exitError, ok := err.(*exec.ExitError); ok &&
+			strings.Contains(string(exitError.Stderr), "Device or resource busy") &&
+			vol.CSIStatus == apiV1.Published {
```
safronovD, line 237: Could we just check that device is really mounted when we see `"Device or resource busy"` error and not operate with volume statuses? In that case we can make error check a part of `SearchPartName` function

libzhang: thanks @safronovD, @yurkov-anton  for the suggestions. Yes,  we can check the device mounted or not to avoid using volume status in SearchPartName.  
here is my questions,  for raw block mode  bind mount and file system mount,  do we have API that can tell if it is mounted or not ?  Please suggest, thank you.

libzhang: Hi @safronovD, @yurkov-anton,  updated code based on your latest comments.  Could you help review again?

safronovD: Sorry for delay. We already have the same method to check mount status - https://github.com/dell/csi-baremetal/blob/eccff10da812669bbab4c529c45e66be8c813972/pkg/base/linuxutils/fs/fs_helper.go#L235

safronovD: Do you resolve problem with the raw block mode?

libzhang: Hi @safronovD,  I see we have this similar function in fs_helper.go,  but we can't use it easily in SeachPartName fuction(partition_operations.go),  since It is implemented for WrapFSImpl.  
another reason is it can't meet our requirement,   since what we used there, the input parameters are something like "/dev/sdh",  we need to remove the /dev prefix to find it in /proc/self/mountinfo. 

for your question on raw block mode,  I already checked,  the /proc/slef/mountinfo can find similar entry in it.  so this method is enough.

pkg/node/provisioners/utilwrappers/partition_operations.go
```diff
@@ -155,6 +158,25 @@ func (d *PartitionOperationsImpl) ReleasePartition(p Partition) error {
 	return nil
 }
 
+// isMounted judge if device like /dev/sda is mounted or not
+func isMounted(device string) bool {
```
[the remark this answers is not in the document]
libzhang: Hi @rkiselenko,  thanks for your comments,  I already checked on a hdd cluster node.  found below issue,   if we use findmnt,  we need to parse the results and check if the related device is in it or not.  
It seems my current solution can achieve either.  so I prefer to current solution. 

```
# cat /proc/self/mountinfo | grep sdi
3815 2218 0:6 /sdi1 /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-c9f66ea3-69ca-4516-a84a-5f6e53ea43f9/globalmount/dev rw,nosuid,relatime shared:626 - devtmpfs udev rw,size=32771740k,nr_inodes=8192935,mode=755
3816 2176 0:6 /sdi1 /hostroot/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-c9f66ea3-69ca-4516-a84a-5f6e53ea43f9/globalmount/dev rw,nosuid,relatime shared:626 - devtmpfs udev rw,size=32771740k,nr_inodes=8192935,mode=755
#

# findmnt -n /dev/sdi1
# findmnt -n /dev/sdi
#
# findmnt | grep sdi
|   |-/hostroot/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-c9f66ea3-69ca-4516-a84a-5f6e53ea43f9/globalmount/dev                                                                  udev[/sdi1]                                                                                                                                                                                        devtmpfs   rw,nosuid,relatime,size=32771740k,nr_inodes=8192935,mode=755
  |-/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-c9f66ea3-69ca-4516-a84a-5f6e53ea43f9/globalmount/dev                                                                             udev[/sdi1]                                                                                                                                                                                        devtmpfs   rw,nosuid,relatime,size=32771740k,nr_inodes=8192935,mode=755

```

rkiselenko (approved):

pkg/node/provisioners/drive_provisioner.go
libzhang: Hi @safronovD,  @yurkov-anton,  may you help take a look at my above comments ?

libzhang: Hi @safronovD,  @yurkov-anton,  sorry to bother you again,  could you please take a look and give any comments or approve the PR ?  Thanks.

mishoyama:
@libzhang please address issue with the DCO check https://github.com/dell/csi-baremetal/pull/887/checks?check_run_id=7386149573

mishoyama:
@libzhang PR validation also failed - https://github.com/dell/csi-baremetal/runs/7386150446?check_suite_focus=true

pkg/node/provisioners/drive_provisioner.go
yurkov-anton: 1. Need to add WrapFSImpl to [PartitionOperations](https://github.com/libzhang/csi-baremetal/blob/620565c880d9a0485f6283f0c2008f3be45040f8/pkg/node/provisioners/utilwrappers/partition_operations.go#L43)
2. I'm not sure need to remove the `/dev` part
```
cat /proc/self/mountinfo | grep /dev/sda
132 99 8:1 / /boot rw,relatime shared:72 - ext4 /dev/sda1 rw,seclabel
```

libzhang:
Hi @mishoyama,  sorry for the late response since I took on vacation last week. 
I updated the DCO and also fixed the go lint errors.  Please take a look.

mishoyama:
@libzhang now codecov/patch check has failed

pkg/node/provisioners/utilwrappers/partition_operations.go
```diff
@@ -155,6 +158,25 @@ func (d *PartitionOperationsImpl) ReleasePartition(p Partition) error {
...
+	// remove the prefix /dev as the mountinfo only contains suffix.
+	device = strings.Trim(device, "/dev")
+	procMounts, err := util.ConsistentRead("/proc/self/mountinfo", 5, time.Millisecond)
```
libzhang, line 165: Hi @mishoyama,  do you have suggestions for this kinds of test coverage ?  since it is reading a file in host,  I could not mock the contents of this file,  so test coverage can't be 100% covered I think.

pkg/node/provisioners/drive_provisioner.go
libzhang: Hi @yurkov-anton, sorry for missing your last comments. Here is why I propose to remove the /dev part,  since I checked on a hdd cluster,  it shows below results,  there is no /dev prefix here,  so we need to cover all scenario here,  remove the prefix is more robust.
```
# cat /proc/self/mountinfo | grep sdi
3815 2218 0:6 /sdi1 /var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-c9f66ea3-69ca-4516-a84a-5f6e53ea43f9/globalmount/dev rw,nosuid,relatime shared:626 - devtmpfs udev rw,size=32771740k,nr_inodes=8192935,mode=755
3816 2176 0:6 /sdi1 /hostroot/var/lib/kubelet/plugins/kubernetes.io/csi/pv/pvc-c9f66ea3-69ca-4516-a84a-5f6e53ea43f9/globalmount/dev rw,nosuid,relatime shared:626 - devtmpfs udev rw,size=32771740k,nr_inodes=8192935,mode=755
#
```

libzhang: and another nvme cluster,  it shows below,
```
admin@provo-rust:~> cat /proc/self/mountinfo  | grep nvme1n1
8941 127 0:5 /nvme1n1 /var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/staging/pvc-644c98ad-9013-4db7-8392-9ad057f6fddb/dev rw,nosuid,noexec shared:1532 - devtmpfs devtmpfs rw,size=4096k,nr_inodes=1048576,mode=755
8956 127 0:5 /nvme1n1 /var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/publish/pvc-644c98ad-9013-4db7-8392-9ad057f6fddb/a786470e-098d-4dfa-957b-94742b0c6925 rw,nosuid,noexec shared:1532 - devtmpfs devtmpfs rw,size=4096k,nr_inodes=1048576,mode=755
9184 127 0:5 /nvme1n1 /var/lib/kubelet/plugins/kubernetes.io/csi/volumeDevices/pvc-644c98ad-9013-4db7-8392-9ad057f6fddb/dev/a786470e-098d-4dfa-957b-94742b0c6925 rw shared:1532 - devtmpfs devtmpfs rw,size=4096k,nr_inodes=1048576,mode=755

```

pkg/node/provisioners/utilwrappers/partition_operations.go
libzhang: add test for this function,  reading host /proc/self/mountinfo,  as any machine should contain this file. 
mock a failure and success case here.   
for success case,  /proc entry should included in every linux machine, so it will return true. 
and for failure case,  we use /dev/nvme50n1,  there should not have such device generally.

pkg/node/provisioners/drive_provisioner.go
yurkov-anton: Hi @libzhang, I got it. Then I think you need remove `/dev` part before call `isMounted`. If both of check check are exist, there will be a conflict. Check with name `isMounted` must check path in mount proc. This check doesn't know about what this path is mean, path to dir or to drive. If you want you can create check for drive, which will be contained removing `/dev` part and `isMounted` call.

libzhang: Hi @yurkov-anton,  sorry,  I did not fully understand your suggestions,  do you mean below ?  
1) remove "/dev" before call isMounted
2) using the WrapFsImpl   implemented IsMounted function?

libzhang: @yurkov-anton,  I updated diff based on above my understanding.

pkg/node/provisioners/utilwrappers/partition_operations.go
```diff
@@ -175,9 +197,15 @@ func (d *PartitionOperationsImpl) SearchPartName(device, partUUID string) (strin
 		// sync partition table
 		err = d.SyncPartitionTable(device)
 		if err != nil {
-			// log and ignore error
-			ll.Errorf("Unable to sync partition table for device %s: %v", device, err)
-			return "", err
+			if exitError, ok := err.(*exec.ExitError); ok &&
```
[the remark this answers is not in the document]
libzhang: @mishoyama , I find a way to increase this test coverage, will update diff later.

yurkov-anton (approved):

safronovD (approved):

pkg/node/provisioners/drive_provisioner.go
yurkov-anton: Yep, it's ok)

pkg/base/linuxutils/partitionhelper/partition_helper.go
```diff
@@ -275,17 +275,17 @@ func (p *WrapPartitionImpl) GetPartitionUUID(device, partNum string) (string, er
 // SyncPartitionTable syncs partition table for specific device
 // Receives device path to sync with partprobe, device could be an empty string (sync for all devices in the system)
 // Returns error if something went wrong
-func (p *WrapPartitionImpl) SyncPartitionTable(device string) (err error) {
+func (p *WrapPartitionImpl) SyncPartitionTable(device string) (outStr string, errStr string, err error) {
```
yurkov-anton, line 278: are you sure that we need to return both of messages? Is outStr used anywhere?

libzhang: currently we have not used the outStr,  but as SyncPartitionTable is using RunCmd (which return the stdout and stderr, err),  we'd better now make it return all.  In case for future usage.

PR #903, opened 2022-07-08
[ISSUE-902] Proposal for implementing storage capacity feature

### LINKED ISSUE #902: Proposal for implementing storage capacity feature
Currently the Plugin uses AC and ACR CRD algoritm for dynamic provisioning, the storage capacity feature is a way to standardize storage tracking mechanism and also may increase performance.

Come up with the proposal for implementing storage capacity feature.

Use custom scheduler, use ACR and extender together.

**Additional context**
N/A

aaletov:
### Resolves #902 

Add proposal for implementing storage capacity feature

## PR checklist
- [x] Add link to the issue
- [x] Choose Project
- [ ] Choose PR label
- [ ] New unit tests added

## Testing


docs/proposals/storage-capacity-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+  // The smallest size that may be used in a
+  // CreateVolumeRequest.capacity_range.limit_bytes field
+  // to create a volume with the same parameters as those in
+  // GetCapacityRequest.
+  MinimumVolumeSize    *wrappers.Int64Value
+}
+```
+
+For each segment and each storage class, CSI GetCapacity is called once with the topology of the segment and the parameters of the class. For each request we should list all
+AC in cluster, filter out AC that do not satisfy segment (node), parameters and volume capacilities of request and find total size of remaining AC for `AvailableCapacity`, max AC
```
mishoyama, line 116: How will that work for LVM storage classes? Will you count non LVG AC? Do we need to remove AC which are reserved when working together with scheduler extender?

docs/proposals/storage-capacity-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+Advantages:
+- First of all, it is a step towards standartization of storage tracking algorithm
+- Implementing Storage Capacity feature may speed up scheduling
+- Generic ephemeral volumes are already supported by driver, they can be provisioned using storage capacity tracking too
+- Storage Capacity objects are being produced and updated independently from PVC. This means that when the pod first goes through the scheduling cycle, it is likely that information about available capacity have already been gathered to Storage Capacity objects and the
+pod can be scheduled at the same cycle
+
+Disadvantages:
+- Storage Capacity objects require more memory in etcd. In contrast, ACR are deleted after volume creation
+- The feature doesn't solve races issues, two pods still may reserve same disk
```
mishoyama, line 32: Will pod stuck in pending state in this case?

docs/proposals/storage-capacity-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+
+| ID      |     Name     | Descriptions | Comments |
+|---------|--------------|--------------|----------|
+| ASSUM-1 |  Performance | Scheduling time may decrease             |  Decreasing of scheduling time was noted in test Kind cluster with disabled Extender |
+
+## Open issues (if applicable)
+
+| ID      | Name | Descriptions | Status | Comments |
+|---------|------|--------------|--------|----------|
+| ISSUE-1 |      |              |        |          |
```
mishoyama, line 132: https://kubernetes.io/docs/concepts/storage/storage-capacity/#limitations 

> One situation where scheduling can fail permanently is when a Pod uses multiple volumes: one volume might have been created already in a topology segment which then does not have enough capacity left for another volume. Manual intervention is necessary to recover from this, for example by increasing capacity or deleting the volume that was already created.

This must be mentioned as open issue

docs/proposals/storage-capacity-proposal.md
aaletov: > How will that work for LVM storage classes?
For LVG SC Volumes capacityplanner now tries to select LVG AC and, in case they have been already reserved or their capacity is to small, tries non-LVG AC. So I suppose that we need to count non-LVG AC, because the volume may be placed on them too. Then MaximumVolumeSize would be max size from (LVG AC + non-LVG AC) and MinimumVolumeSize would be some constant threshold.

There I have a question. You use AcSizeMinThresholdBytes (1MB) as the minimum size of AC after which it must be deleted, but minimum Volume size for LVG is determined by DefaultPESize and equals 4MB. Which is the minimum LVG Volume size for plugin?

aaletov: > Do we need to remove AC which are reserved when working together with scheduler extender?

I suppose we should filter out reserved AC in case of non-LVG SC and subtract total reservation size from AC in case of LVG AC. There is no easy way to check reservations for AC, so we should list all ACR from cluster and loop over them, or add information about reservations in AC CRD

docs/proposals/storage-capacity-proposal.md
aaletov: Added issue

docs/proposals/storage-capacity-proposal.md
aaletov: Honestly, I don't know how this happens, @safronovD have mentioned this issue, but he's busy now

safronovD: Or user/Operator should remove PVC and delete pod to restart scheduling process as workaround

docs/proposals/storage-capacity-proposal.md
```diff
@@ -0,0 +1,132 @@
...
+  # StatefulSet (only needs to get Pod) and Deployment (needs to get
+  # Pod and then ReplicaSet to find the Deployment).
+  - apiGroups: [""]
+    resources: ["pods"]
+    verbs: ["get"]
+  - apiGroups: ["apps"]
+    resources: ["replicasets"]
+    verbs: ["get"]
+```
+To enable using these object during scheduling `storageCapacity: true` must be added to CSIDriver spec.
```
safronovD, line 81: Is this field mutable? Could user enable/disable this feature without deleting CSI?

aaletov: This field was immutable in Kubernetes <= 1.22 and mutable since 1.23 

https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.24/#csidriver-v1-storage-k8s-io

aaletov: Added information about field's mutability

docs/proposals/storage-capacity-proposal.md
aaletov: @mishoyama I think that we should not consider reservations at all. The goal of proposed changes is to reduce number of nodes in filter() request (which is not obvious from proposal though). Also, we can't guarantee correct pod placement without ACR, so no changes required in this mechanism. For that goal, it is better if wrong nodes will be included in request, than if right nodes will be filtered out. For example, if we have 2 nodes with 1 drive on each, which are reserved for some pending pod, both nodes must be considered by extender for second pod, because it is actually feasible, we just need to wait until first pod will be placed and ACR will be deleted. Same logic is applied to LVM SC, we must count all AC, because only primary rough filtering is required from default scheduler, the main work does the extender

mishoyama (approved):

PR #927, opened 2022-11-28
[ISSUE-926] Change Pods Max Reschedule Interval in k8s 1.24 and 1.25 

### LINKED ISSUE #926: It Takes Too Much Time for CSI to provision PVCs on Kubernetes 1.24
**Describe the bug**
It Takes Too Much Time for CSI to provision PVCs on Kubernetes 1.24. In many cases, the provision time of 1 pod with CSI-managed PVCs increases to over 5 minutes. As a result, some applications based on this baremetal CSI with many pods with CSI-managed PVCs will totally spend over 1 hour on the PVCs provision during applications deployment.

 - Hardware: Any
 - OS: Any
 - Version: Kubernetes 1.24

**To Reproduce**
You can see this long provision time on any kubernetes 1.24 with CSI installed and PVCs provision managed by CSI.

**Expected behavior**
CSI-managed PVCs provision time should be reduced to reasonable time.

CraneShiEMC:
Signed-off-by: Shi, Crane <crane.shi@emc.com>

### Resolves #926 

Change Pods Max Reschedule Interval to 60s via scheduler extender patcher in k8s 1.24 and 1.25. This change only applies to rke2 platform.

## PR checklist
- [ ] Add link to the issue
- [ ] Choose Project
- [ ] Choose PR label
- [ ] New unit tests added

## Testing
On an Atlantic cluster whose k8s version is 1.24, with scheduler extender patch image including this change, the provision time of 1 pod with CSI-managed PVCs has been reduced to about 1min as expected and the total provsion time of multiple pods with CSI-managed PVCs has also been reduced a lot.
On another Atlantic cluster whose k8s version is 1.21, with scheduler extender patch image including this change, this parameter has not been added to kube-scheduler init parameter as expected.

### DIFF
--- a/pkg/scheduler/patcher/main.py
+++ b/pkg/scheduler/patcher/main.py
@@ -109,12 +109,15 @@ def run():
     if kube_major_ver==1 and kube_minor_ver>22:
         path = args.target_config_23_path
 
+    config_unschedulable_pods = (kube_major_ver==1 and kube_minor_ver>23 and kube_minor_ver<26)
+
     manifest = "/etc/kubernetes/manifests/kube-scheduler.yaml"
     if args.platform == "rke":
         manifest = "/var/lib/rancher/rke2/agent/pod-manifests/kube-scheduler.yaml"
 
     manifest = ManifestFile(
-        manifest, [config_volume, policy_volume, config_19_volume, config_23_volume], path, args.backup_path)
+        manifest, [config_volume, policy_volume, config_19_volume, config_23_volume],
+        path, args.backup_path, config_unschedulable_pods)
 
     # add watcher on signals
     killer = GracefulKiller(args.restore, manifest)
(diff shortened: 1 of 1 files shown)


yanxiucai (commented):
Suggest to investigate why Kubernetes 1.24 changed Pod Max Reschedule Interval from 60s to 5 mins first, then we can make a decision whether to commit this change.

CraneShiEMC:
> Suggest to investigate why Kubernetes 1.24 changed Pod Max Reschedule Interval from 60s to 5 mins first, then we can make a decision whether to commit this change.

The original intention of the change of “rescheduling interval” from 60s to 5mins is to alleviate the pods starvation issue in scheduling pod in large distributed systems, with thousands of pods or more, and Kubernetes community plans to remove this hard limit of “rescheduling interval” in the future release. 
The pods starvation issue in scheduling pod is that the continual rescheduling of unschedulable higher-priority pods blocks the normal scheduling of schedulable lower-priority pods. 
In the scenario of large-scale system, there might be huge numbers of higher-priority unschedulable pods. If we use 60s “max time of pod rescheduling interval”, i.e. each of those higher-priority unschedulable pods will be rescheduled at most 60s later and even less, there might be unschedulable higher-priority pods continually chosen to try re-scheduling (which will still fail), which will keep blocking the normal scheduling of actually schedulable pods with lower priority. 
This issue has little influence on scheduling in the normal situation.

To sum up, the side effect of changing the “max time of pod rescheduling interval” from 5mins back to 60s may be the increased possibility of occurrence of pods starvation issue in large-scale deployment.

libzhang (approved):

lordjun (approved):

PR #961, opened 2023-02-14
Bugfix issue 959 correct volume cr size

yanxiucai:
Align volumeCR size to MB when it is not LVG SC

### DIFF
--- a/pkg/base/capacityplanner/common.go
+++ b/pkg/base/capacityplanner/common.go
@@ -41,11 +41,11 @@ func AlignSizeByPE(size int64) int64 {
 
 // AlignSizeByMB make size aligned with MB
 func AlignSizeByMB(size int64) int64 {
-        reminder := size % (int64(util.MBYTE))
-        if reminder != 0 {
-                return size -  reminder
-        }
-        return size
+	reminder := size % (int64(util.MBYTE))
+	if reminder != 0 {
+		return size - reminder
+	}
+	return size
 }
 
 // SubtractLVMMetadataSize subtracts LVM metadata size from raw drive size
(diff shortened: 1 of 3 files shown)


pkg/base/capacityplanner/common.go
```diff
@@ -39,6 +39,15 @@ func AlignSizeByPE(size int64) int64 {
 	return size + alignement
 }
 
+// AlignSizeByMB make size aligned with MB
+func AlignSizeByMB(size int64) int64 {
```
libzhang, line 43: please add unit-test in common_test.go for this function.

pkg/common/volume_operations.go
```diff
@@ -241,7 +241,12 @@ func (vo *VolumeOperationsImpl) handleVolumeCreation(ctx context.Context, log *l
 	vo.cache.Set(v.Id, podNamespace)
 
 	// decrease AC size
-	ac.Spec.Size -= allocatedBytes
+        if util.IsStorageClassLVG(sc) {
+                ac.Spec.Size -= allocatedBytes
```
libzhang, line 245: considering such scenario,  LVG group size is not MB align,  then it seems we can never used up the LVG ?  ac.Spec.Size will somehow not zero, right?

yanxiucai: Yes, you are right.  LVG group ac.Spec.Size  is not MB aligned, while allocatedBytes is 4MB aligned 
https://github.com/dell/csi-baremetal/blob/master/pkg/common/volume_operations.go#L206

pkg/base/capacityplanner/common.go
yanxiucai: added

CraneShiEMC (approved):

libzhang (approved):

yanxiucai:
Passed both the csi-custom-ci and csi-custom-acceptance tests

PR #992, opened 2023-04-17
[ISSUE-991]  disk taint-like mechanism

### LINKED ISSUE #991: A method to mark a disk drive unschedulable is desired

As a CSI consumer, I hope it can provide a method to taint a disk drive and mark it unschedulable, possibly by using a label. With the disk marked, no new PV will be allocated from the drive, but it won't affect existing PVs.

This may provide a choice for applications when they encounter disk read/write errors. There might be a time gap between application hitting a disk error and CSI disk replacement workflow taking place. If an application can only wait for CSI disk replacement, it may need to panic on disk error situation repeatedly for a while, or use fake-attach method but with the risk of losing data. The application may want to proactively restart with new PVs scheduled on other healthy drives. So it needs to avoid being scheduled to the same drive.

 Acceptance Criteria:

feature usage is documented.
implementation is done in which a disk can be marked as unschedulable, no new PVs will be allocated from this disk, but existing PVs won't be affected.
the implementation conforms to [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md)

I hope it can provide a method to taint a disk drive and mark it unschedulable, possibly by using a label. With the disk marked, no new PV will be allocated from the drive, but it won't affect existing PVs.

None.

**Additional context**
None.

libzhang:
### Resolves #991
move code to master branch.

## PR checklist
- [x] Add link to the issue
- [ ] Choose Project
- [ ] Choose PR label
- [x] New unit tests added

## Testing
_Provide test details_

### DIFF
--- a/pkg/crcontrollers/drive/drivecontroller.go
+++ b/pkg/crcontrollers/drive/drivecontroller.go
@@ -475,9 +475,7 @@ func (c *Controller) handleDriveLableUpdate(ctx context.Context, log *logrus.Ent
 	location := drive.GetName()
 	lvg, err := c.crHelper.GetLVGByDrive(ctx, drive.Spec.UUID)
 	if err != nil {
-		if err != errTypes.ErrorNotFound {
-			log.Infof("LVG for drive %s is not found: %s", drive.GetName(), err.Error())
-		}
+		log.Infof("LVG for drive %s is not found: %s", drive.GetName(), err.Error())
 	} else {
 		if lvg != nil {
 			location = lvg.GetName()
--- a/pkg/crcontrollers/drive/drivecontroller.go
+++ b/pkg/crcontrollers/drive/drivecontroller.go
@@ -140,6 +140,9 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 }
 
 func (c *Controller) handleDriveUpdate(ctx context.Context, log *logrus.Entry, drive *drivecrd.Drive) (uint8, error) {
+	// handle drive labeled or not
+	if err := c.handleDriveLableUpdate(ctx, log, drive); err != nil {
+	}
 	// handle offline/online drive status
 	if err := c.handleDriveStatus(ctx, drive); err != nil {
 		return ignore, err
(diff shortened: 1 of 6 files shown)


pkg/crcontrollers/drive/drivecontroller.go
```diff
@@ -140,6 +140,9 @@ func (c *Controller) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Resu
 }
 
 func (c *Controller) handleDriveUpdate(ctx context.Context, log *logrus.Entry, drive *drivecrd.Drive) (uint8, error) {
+	// handle drive labeled or not
+	if err := c.handleDriveLableUpdate(ctx, log, drive); err != nil {
```
yimingwangdell, line 144: maybe better to deal with err

pkg/crcontrollers/drive/drivecontroller.go
```diff
@@ -455,3 +458,59 @@ func (c *Controller) removeRelatedAC(ctx context.Context, log *logrus.Entry, cur
...
+	if acLabels == nil {
+		// label is not existed on CR, just create it and attach to CR
+		acLabels = map[string]string{}
+		ac.Labels = acLabels
+	}
+	if taintLabelExisted {
+		acLabels[apiV1.DriveTaintKey] = apiV1.DriveTaintValue
+	} else {
+		if _, ok := acLabels[apiV1.DriveTaintKey]; ok {
+			delete(acLabels, apiV1.DriveTaintKey)
```
yimingwangdell, line 506: I think delete can handle non-existing case

pkg/crcontrollers/drive/drivecontroller.go
libzhang: This is intentional here now.  now label update should not block other functions.  and the err is also logged in function handleDriveLableUpdate.

pkg/crcontrollers/drive/drivecontroller.go
libzhang: Good catch. I checked the golang doc, will remove such check here. 

>The delete built-in function deletes the element with the specified key (m[key]) from the map. If m is nil or there is no such element, delete is a no-op.

pkg/crcontrollers/drive/drivecontroller.go
```diff
@@ -455,3 +459,53 @@ func (c *Controller) removeRelatedAC(ctx context.Context, log *logrus.Entry, cur
...
+		// label is not existed on CR, just create it and attach to CR
+		acLabels = map[string]string{}
+		ac.Labels = acLabels
+	}
+	if taintLabelExisted {
+		acLabels[apiV1.DriveTaintKey] = apiV1.DriveTaintValue
+	} else {
+		delete(acLabels, apiV1.DriveTaintKey)
+	}
+
```
CraneShiEMC, line 504: Do we also need to sync this label to the corresponding lvg object for potential use?

CraneShiEMC (approved):

pkg/crcontrollers/drive/drivecontroller.go
libzhang: currently, no sync to lvg object as there is no case to use lvg to schedule.  we may add it when it is necessary.

yimingwangdell (approved):

aiqiEMC (approved):

PR #997, opened 2023-05-05
[ISSUE-1018] Custom Storage Group Feature

### LINKED ISSUE #1018: Custom Storage Group Feature
I need Bare-metal CSI driver to allow selecting a specific device for some Persistence Volume (PV)

More specifically, the situations to support are:
1. one pod will be able to choose a specific drive for its PV when there are multiple drives of the same storage type (storage type means HDD/SSD/NVME etc., can be LVG class)
2. one pod using multiple PVs will be able to distribute the PVs on different drives of the same storage type for better throughput

Acceptance criteria:
a) The implementation conforms to [CSI spec](https://github.com/container-storage-interface/spec/blob/master/spec.md)
b) Feature usage is documented on the CSI page
c) One pod will be able to choose a specific drive for its PV when there are multiple drives of the same storage type (storage type means HDD/SSD/NVME etc., can be LVG class)
d) One pod using multiple PVs will be able to distribute the PVs on different drives of the same storage type for better throughput

I solve the problem by this custom storage group approach.

In this approach, user can create custom storage group, i.e. a new Custom Resource StorageGroup of Bare-metal CSI, and add specific drives into the custom storage group.

Then, the corresponding custom-storage-group k8s label, drive.csi-baremetal.dell.com/storage-group: ${storage-group-name}, would be immediately added to all the CSI drives selected in this storage group by storagegroup controller after each custom storage group created.

Then, to make PVCs land on those specific drives, just add the corresponding custom storage group label drive.csi-baremetal.dell.com/storage-group: ${storage-group-name} in volume claim spec

Custom Storage Class
Disk affinity/anti-affinity

**Additional context**
None

CraneShiEMC:
### Resolves #1018 

Support the custom storage group feature.

## PR checklist
- [ ] Add link to the issue
- [ ] Choose Project
- [ ] Choose PR label
- [ ] New unit tests added

## Testing
The custom storage group feature has functioned as expected in our various tests, including normal function and service procedure support.
Passed CSI custom ci: https://asd-ecs-jenkins.isus.emc.com/job/csi-custom-ci/1549/


pkg/base/capacityplanner/node_capacity.go
```diff
@@ -154,7 +154,7 @@ func (nc *nodeCapacity) selectACForVolume(vol *genV1.Volume) *accrd.AvailableCap
 	}
 
 	for _, ac := range nc.acsOrder[vol.StorageClass] {
-		if requiredSize <= nc.acs[ac].Spec.Size {
+		if requiredSize <= nc.acs[ac].Spec.Size && nc.acs[ac].Labels[v1.StorageGroupLabelKey] == vol.StorageGroup {
```
libzhang, line 157: I think for this,  you may pre-filter acs as https://github.com/dell/csi-baremetal/blob/master/pkg/base/capacityplanner/node_capacity.go#L75

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
```diff
@@ -0,0 +1,586 @@
...
+		Complete(c)
+}
+
+func (c *Controller) filterUpdateEvent(old runtime.Object, new runtime.Object) bool {
+	var (
+		oldDrive *drivecrd.Drive
+		newDrive *drivecrd.Drive
+		ok       bool
+	)
+	if oldDrive, ok = old.(*drivecrd.Drive); !ok {
```
libzhang, line 77: I don't think this line code is necessary here.  we can remove it.

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
```diff
@@ -0,0 +1,586 @@
...
+	if sgStatus == apiV1.Creating {
+		return c.handleStorageGroupCreation(ctx, log, storageGroup)
+	}
+
+	return ctrl.Result{}, nil
+}
+
+// combine the following similar funcs
+func (c *Controller) removeACStorageGroupLabel(ctx context.Context, log *logrus.Entry, ac *accrd.AvailableCapacity) error {
+	delete(ac.Labels, apiV1.StorageGroupLabelKey)
```
libzhang, line 160: is any chance ac.Labels is nil ?

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
```diff
@@ -0,0 +1,586 @@
...
+		case "int64":
+			fieldValueInt64, _ := strconv.ParseInt(fieldValue, 10, 64)
+			if driveField.Int() != fieldValueInt64 {
+				return false
+			}
+		case "bool":
+			fieldValueBool, _ := strconv.ParseBool(fieldValue)
+			if driveField.Bool() != fieldValueBool {
+				return false
+			}
```
libzhang, line 485: here should we add default case here for un-intention type ?

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
```diff
@@ -0,0 +1,586 @@
...
+	ctx, cancelFn := context.WithTimeout(ctx, contextTimeoutSeconds*time.Second)
+	defer cancelFn()
+
+	// read name
+	name := req.Name
+	// customize logging
+	log := c.log.WithFields(logrus.Fields{"method": "Reconcile", "name": name})
+
+	drive := &drivecrd.Drive{}
+	if err := c.client.ReadCR(ctx, name, "", drive); err == nil {
```
yimingwangdell, line 102: need to add log info for error

pkg/base/capacityplanner/node_capacity.go
CraneShiEMC: Hi Libin, it is different from the case of your disk taint feature that you can do the ACs pre-filter directly based on whether the AC has disk taint label in this initial nodeCapacity construction stage. 
In storage group scenario, it's even a common case that an ACR contains multiple volume requests to multiple storage groups resulting that we can only do the ACs filtering on different storage group for each volume request at the stage of selecting AC for the volume request, i.e. the place at selectACForVolume func you see.

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
CraneShiEMC: Hi Libin, at present, the storagegroup controller watches the events of both drive objects and storagegroup objects, so this code here is to distinguish the non-drive object, i.e. storagegroup object. Without this line, we will miss the storagegroup update event and then the storage group reconcilation wouldn't be normally handled. Though the StorageGroupSpec update is not supported, we actually still do the necessary update on the whole StorageGroup object during the storagegroup reconcilation for storagegroup status tracking, finalizer addition or removal, node addtion support etc. Therefore, the catch of storagegroup object update event is still required here.
Here, I just use the similar code logic to those already used in other CR controllers watching object events of multiple types.

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
CraneShiEMC: In most cases, AC will have CSI default labels. But I will handle the case you mentioned.

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
CraneShiEMC: I enumerate all the drive CR's property types here. Ok, I will change the code to add default case for un-intention type according to your comment.

Makefile
```diff
@@ -120,6 +121,7 @@ generate-baremetal-crds: install-controller-gen
 	controller-gen $(CRD_OPTIONS) paths=api/v1/drivecrd/drive_types.go paths=api/v1/drivecrd/groupversion_info.go output:crd:dir=$(CSI_CHART_CRDS_PATH)
 	controller-gen $(CRD_OPTIONS) paths=api/v1/lvgcrd/logicalvolumegroup_types.go paths=api/v1/lvgcrd/groupversion_info.go output:crd:dir=$(CSI_CHART_CRDS_PATH)
 	controller-gen $(CRD_OPTIONS) paths=api/v1/nodecrd/node_types.go paths=api/v1/nodecrd/groupversion_info.go output:crd:dir=$(CSI_CHART_CRDS_PATH)
+	controller-gen $(CRD_OPTIONS) paths=api/v1/storagegroupcrd/storagegroup_types.go paths=api/v1/storagegroupcrd/groupversion_info.go output:crd:dir=$(CSI_CHART_CRDS_PATH)
```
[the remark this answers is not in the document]
CraneShiEMC: I've actually had a PR on csi-baremetal-operator for the corresponding storage group feature including the addition of sg crd, https://github.com/dell/csi-baremetal-operator/pull/180... Sorry for forgetting to provide it before...

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
libzhang: yes, I originally thought you can "return true" in the end(Line 84)  to handle other case.  anyway,  this is ok.

CraneShiEMC: @libzhang Very sorry for misunderstanding of your original comments. 
Your suggested code change is much better if I want to catch all storagegroup update events and only storagegroup label update event for drive objects.
I will change the code here accordingly.

pkg/base/capacityplanner/node_capacity.go
libzhang: ok,  then below logic,  what if no storageGroup acs?   or every ac have a default storageGroup? 
```golang
if requiredSize <= nc.acs[ac].Spec.Size && nc.acs[ac].Labels[v1.StorageGroupLabelKey] == vol.StorageGroup
```

CraneShiEMC: No-storagegroup ACs can still work here to match the no-storagegroup volume requests (and only match no-storagegroup volume requests) here.
Those ACs without storagegroup label, i.e. no-storagegroup ACs, return "" (empty string) for nc.acs[ac].Labels[v1.StorageGroupLabelKey] while the volume requests without storage group also return "" (empty string) for vol.StorageGroup. Therefore, they can also match.

Actually, this "nil storagegroup", can be thought of as some kind of "default storagegroup" you mentioned to some extent.

At present, we have many previous tests cases without storage group can still function well on the csi images with storage group feature.

CraneShiEMC: @libzhang 
No-storagegroup ACs can still work here to match the no-storagegroup volume requests (and only match no-storagegroup volume requests) here.
Those ACs without storagegroup label, i.e. no-storagegroup ACs, return "" (empty string) for nc.acs[ac].Labels[v1.StorageGroupLabelKey] while the volume requests without storage group also return "" (empty string) for vol.StorageGroup. Therefore, they can also match.

Actually, this "nil storagegroup", can be thought of as some kind of "default storagegroup" you mentioned to some extent.

At present, all the previous tests cases without storage group can still function well on the csi images with storage group feature.

CraneShiEMC: @libzhang Has refined the code here to cover the case of AC without any labels. 
(But actually, all CSI-created ACs would contain default CSI app labels. The only very low probability of the case of AC without any labels is for the manually-created AC without labels...)

libzhang (approved):

yimingwangdell (approved):

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
```diff
@@ -0,0 +1,586 @@
...
+					return ctrl.Result{Requeue: true}, err
+				}
+			}
+		}
+	}
+	return ctrl.Result{}, nil
+}
+
+func (c *Controller) syncDriveStorageGroupLabel(ctx context.Context, drive *drivecrd.Drive) (ctrl.Result, error) {
+	log := c.log.WithFields(logrus.Fields{"method": "syncDriveStorageGroupLabel", "name": drive.Name})
```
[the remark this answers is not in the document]
CraneShiEMC: added

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
```diff
@@ -0,0 +1,604 @@
...
+	if err := c.client.ReadCR(ctx, name, "", drive); err == nil {
+		return c.syncDriveStorageGroupLabel(ctx, drive)
+	} else if !k8serrors.IsNotFound(err) {
+		log.Errorf("error in reading %s as drive object: %v", name, err)
+	}
+
+	storageGroup := &sgcrd.StorageGroup{}
+	if err := c.client.ReadCR(ctx, name, "", storageGroup); err != nil {
+		log.Warnf("Failed to read StorageGroup %s with error: %v", name, err)
+		return ctrl.Result{}, client.IgnoreNotFound(err)
```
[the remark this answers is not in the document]
CraneShiEMC: We can delete sg to make the delete event enter the reconcilation queue of the controller even after the object has been deleted. 
On the other hand, the finalizer of sg will block normal deletion of sg until all the necessary clean tasks of this sg finished.
So the NotFound error here will not be a real error.

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
```diff
@@ -0,0 +1,604 @@
...
+	drive := &drivecrd.Drive{}
+	if err := c.client.ReadCR(ctx, name, "", drive); err == nil {
+		return c.syncDriveStorageGroupLabel(ctx, drive)
+	} else if !k8serrors.IsNotFound(err) {
+		log.Errorf("error in reading %s as drive object: %v", name, err)
+	}
+
+	storageGroup := &sgcrd.StorageGroup{}
+	if err := c.client.ReadCR(ctx, name, "", storageGroup); err != nil {
+		log.Warnf("Failed to read StorageGroup %s with error: %v", name, err)
```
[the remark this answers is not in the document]
CraneShiEMC: has changed the log level to errorf

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
```diff
@@ -0,0 +1,604 @@
...
+		storageGroup.Annotations[sgTempStatusAnnotationKey] = apiV1.Creating
+		if err := c.client.UpdateCR(ctx, storageGroup); err != nil {
+			log.Errorf("Unable to update StorageGroup status, error: %v.", err)
+			return ctrl.Result{Requeue: true}, err
+		}
+	}
+
+	if sgStatus == apiV1.Creating {
+		return c.handleStorageGroupCreationOrUpdate(ctx, log, storageGroup)
+	}
```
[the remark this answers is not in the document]
CraneShiEMC: The unexpected sgStatus will not influence the sg handling. At present, we only handle the creation or update of sg in Creating status

pkg/crcontrollers/storagegroup/storagegroupcontroller.go
```diff
@@ -0,0 +1,620 @@
...
+		drive := drive
+		if drive.Labels[apiV1.StorageGroupLabelKey] == sg.Name {
+			if err := c.removeDriveAndACStorageGroupLabel(ctx, log, &drive, sg); err != nil {
+				log.Errorf("error in remove storage-group label from drive %s", err.Error())
+				removalNoError = false
+			}
+		}
+	}
+	if !removalNoError {
+		return ctrl.Result{Requeue: true}, fmt.Errorf("error in removing storage-group label")
```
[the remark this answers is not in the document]
CraneShiEMC: Here, we have multiple drives and ACs for labeling removal. If we have multiple errors here, we cannot only return one of the multiple removal errs.

To avoid the err messages missing, I've just refined the logic to combine all the possible removal errs into one error to be returned if we encounter any error in the labeling removal of multiple drives and ACs.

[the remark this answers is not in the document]
yimingwangdell: yes, I got you, but I mean replace removalNoError:=true with removalError:=false, and change !removalNoError(no no error) to removalError(err) maybe better.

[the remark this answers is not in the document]
CraneShiEMC: ok, your suggestion is better. Sorry for the misunderstanding...

But I just found that, with the introduction of labelRemovalErrMsgs to refine error messages for label removal, the bool flag removalNoError or removalErr can be deprecated... We can just use the length of labelRemovalErrMsgs to judge whether there is removalErr

yimingwangdell (approved):

yimingwangdell (approved):

PR #1050, opened 2023-08-26
Need to Retry on Failed Volume Also in NodePublish/NodeUnpublish/NodeUnstage

CraneShiEMC:
### Resolves #

Need to Retry on Failed Volume Also in NodePublish/NodeUnpublish/NodeUnstage

## PR checklist
- [ ] Add link to the issue
- [ ] Choose Project
- [ ] Choose PR label
- [ ] New unit tests added

## Testing
_Provide test details_

### DIFF
--- a/pkg/node/node.go
+++ b/pkg/node/node.go
@@ -353,7 +353,9 @@ func (s *CSINodeService) NodeUnstageVolume(ctx context.Context, req *csi.NodeUns
 	}
 
 	currStatus := volumeCR.Spec.CSIStatus
-	if currStatus == apiV1.Created {
+	if currStatus == apiV1.Failed {
+		ll.Warningf("Volume status: %s. Need to retry.", currStatus)
+	} else if currStatus == apiV1.Created {
 		ll.Info("Volume has been already unstaged")
 		return &csi.NodeUnstageVolumeResponse{}, nil
 	} else if currStatus != apiV1.VolumeReady {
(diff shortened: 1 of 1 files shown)


pkg/node/node.go
```diff
@@ -488,6 +494,27 @@ func (s *CSINodeService) NodePublishVolume(ctx context.Context, req *csi.NodePub
 			resp, errToReturn = nil, fmt.Errorf("failed to publish volume: fake attach error %s", err.Error())
 		}
 	} else {
+		// will check whether srcPath is mounted, if not, need to redo NodeStageVolume
```
libzhang, line 497: I am a little concerned about this process,  as some mount logic(like global path setup) is done in kubelet.  there redo nodeStage may not help in such case.  do we have a test can prove that this will help in the failure case?

CraneShiEMC: The kubelet's wrong CSI call NodePublishVolume in this case assumed that the volume's device global path's mountpoint has been successfully setup in some successful previous CSI call NodeStageVolume. But acutally the volume's device global path has been unmounted in the forceful-node-removal (possibly) and kubelet cannot successfully sync the volume's real status because of its continually-failed "orphan" pod's volume cleanup by CSI call NodeUnpublish. 

In spite of this, the device global path still exists there if the volume still exists in the most cases. Even though the device global path has also been removed in the worst case, CSI current logic can still create the device global path itself.

I will do the functional test on this code change to verify it.

libzhang:
it seems the UT  in PR validation failed. please fix it.

lordjun (approved):

pkg/node/node.go
```diff
@@ -602,6 +649,9 @@ func (s *CSINodeService) NodeUnpublishVolume(ctx context.Context, req *csi.NodeU
 	volumeCR.Spec.Owners = owners
 	if len(volumeCR.Spec.Owners) == 0 {
 		volumeCR.Spec.CSIStatus = apiV1.VolumeReady
+	} else {
```
libzhang, line 652: it seems this is not needed,  the default state is Published in this function.

libzhang (approved):

pkg/node/node.go
CraneShiEMC: because we will proceed on the failed volume now, here the reset of status to Published is for Failed volume successfully unpublished and used by multiple pods.

pkg/node/node.go
CraneShiEMC: I've similuated the scenario of volume's k8s global device mountpoint missing in my standalone test. This CSI defensive enhancement can work well as expected in the test.

PR #1094, opened 2024-02-24
Finding drive vid in lsblk vendor

### LINKED ISSUE #1095: Value of disk vendor id differs between HAL and output from lsblk.
**Describe the bug**

 - Hardware: [e.g. Dell PowerEdge]
 - OS: [e.g. Ubuntu, run `printf "$(uname -srm)\n$(cat /etc/os-release)\n"`]
 - Version: [e.g. 1.0.0]

**To Reproduce**
Steps to reproduce the behavior.

**Expected behavior**

**Screenshots**

**Additional context**

prasant123:
## PR checklist
- [x] Add link to the issue
- [ ] Choose Project
- [ ] Choose PR label
- [ ] New unit tests added

## Testing
_Provide test details_

### DIFF
--- a/pkg/base/linuxutils/lsblk/lsblk.go
+++ b/pkg/base/linuxutils/lsblk/lsblk.go
@@ -173,8 +173,9 @@ func (l *LSBLK) SearchDrivePath(drive *api.Drive) (string, error) {
 	vid := drive.VID
 	pid := drive.PID
 	for _, l := range lsblkOut {
-		if strings.EqualFold(l.Serial, sn) && strings.Contains(l.Vendor, vid) &&
-			strings.EqualFold(l.Model, pid) {
+		lvid := strings.TrimSpace(l.Vendor)
+		if strings.EqualFold(l.Serial, sn) && strings.EqualFold(lvid, vid) &&
+			strings.HasPrefix(l.Model, pid) {
 			device = l.Name
 			break
 		}
--- a/pkg/base/linuxutils/lsblk/lsblk.go
+++ b/pkg/base/linuxutils/lsblk/lsblk.go
@@ -162,6 +162,7 @@ func (l *LSBLK) GetBlockDevices(device string) ([]BlockDevice, error) {
 // Returns drive's path based on provided drivecrd.Drive or error if something went wrong
 func (l *LSBLK) SearchDrivePath(drive *api.Drive) (string, error) {
 
+	device := ""
 	// try to find it with lsblk
 	lsblkOut, err := l.GetBlockDevices("")
 	if err != nil {
--- a/pkg/base/linuxutils/lsblk/lsblk.go
+++ b/pkg/base/linuxutils/lsblk/lsblk.go
@@ -161,11 +161,6 @@ func (l *LSBLK) GetBlockDevices(device string) ([]BlockDevice, error) {
 // Receives an instance of drivecrd.Drive struct
 // Returns drive's path based on provided drivecrd.Drive or error if something went wrong
 func (l *LSBLK) SearchDrivePath(drive *api.Drive) (string, error) {
-	// device path might be already set by hwmgr
-	device := drive.Path
-	if device != "" {
-		return device, nil
-	}
 
 	// try to find it with lsblk
 	lsblkOut, err := l.GetBlockDevices("")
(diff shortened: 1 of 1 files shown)


eddiepavkovic (approved):
I'm good with the change as it was discussed with @mishoyama

pkg/base/linuxutils/lsblk/lsblk.go
```diff
@@ -178,7 +173,7 @@ func (l *LSBLK) SearchDrivePath(drive *api.Drive) (string, error) {
 	vid := drive.VID
 	pid := drive.PID
 	for _, l := range lsblkOut {
-		if strings.EqualFold(l.Serial, sn) && strings.EqualFold(l.Vendor, vid) &&
+		if strings.EqualFold(l.Serial, sn) && strings.Contains(l.Vendor, vid) &&
```
libzhang, line 176: how about the pid handling ?  lsblk returned "HGST HUS728T8TALE6L0",  but the drive CR value is from drivemgr,  which is "HGST HUS728T8TAL". 
and also is there possibility that Serinal need same handling with Vendor,  as Serialnumber may also contain some trailing space.

pkg/base/linuxutils/lsblk/lsblk.go
```diff
@@ -161,11 +161,6 @@ func (l *LSBLK) GetBlockDevices(device string) ([]BlockDevice, error) {
 // Receives an instance of drivecrd.Drive struct
```
libzhang, line 161: one general comments is to create an issue first and link the issue to the PR.  and put [ISSUE-xxx] in PR title.

prasant123: Addressed

pkg/base/linuxutils/lsblk/lsblk.go
prasant123: We are observing failure only for the Vendo ID. Since the serial number reported by both HAL and lsblk are the same, we should be good with the existing implementation. We need a quick resolution due to the time constraint, however I agree that we need to have a relook at this area of code in the next release cycle

libzhang: do you mean for the pid I mentioned above,  for different length strings,   strings.EqualFold will return true in the case ?

pkg/base/linuxutils/lsblk/lsblk.go
```diff
@@ -178,7 +173,7 @@ func (l *LSBLK) SearchDrivePath(drive *api.Drive) (string, error) {
 	vid := drive.VID
 	pid := drive.PID
 	for _, l := range lsblkOut {
-		if strings.EqualFold(l.Serial, sn) && strings.EqualFold(l.Vendor, vid) &&
+		if strings.EqualFold(l.Serial, sn) && strings.Contains(l.Vendor, vid) &&
 			strings.EqualFold(l.Model, pid) {
```
yanxiucai, line 177: Need code change here to handle the PID.
In our test, for the same disk, the VID and PID format is different from lsblk and drive CR. 

 l.Vendor  is  "ATA     " while  vid (drive.VID) is “ATA”
l.Model  is "HGST HUS728T8TALE6L0" while pid (drive.PID) is “HGST HUS728T8TAL”

pkg/base/linuxutils/lsblk/lsblk.go
yanxiucai: I think we should only ignore the trailing space in this JIRA. It is not safe to use "string.contains(A, B)" as it will also return TRUE for many other cases. For example:  A = "testing   "    B="ing"

pkg/base/linuxutils/lsblk/lsblk.go
prasant123: Agree!
Updated the changes, please take a look

pkg/base/linuxutils/lsblk/lsblk.go
```diff
@@ -178,8 +174,9 @@ func (l *LSBLK) SearchDrivePath(drive *api.Drive) (string, error) {
 	vid := drive.VID
 	pid := drive.PID
 	for _, l := range lsblkOut {
-		if strings.EqualFold(l.Serial, sn) && strings.EqualFold(l.Vendor, vid) &&
-			strings.EqualFold(l.Model, pid) {
+		lvid := strings.TrimSpace(l.Vendor)
+		if strings.EqualFold(l.Serial, sn) && strings.EqualFold(lvid, vid) &&
+			strings.HasPrefix(l.Model, pid) {
```
libzhang, line 179: basically, it is ok. 
just noted that: if pid is null string(""),  then strings.HasPrefix will also return true.  but in our case,  pid seems to not able to be null. we may need re-look at this code in later release as well.

lordjun (approved):

lordjun:
Please go ahead to merge the code, Prasant. Thanks.

